PR #18, opened 2020-02-08
Resolve compiler warnings and replace deprecated code

### LINKED ISSUE #16: Resolve compiler warnings and replace deprecated code
At the moment we have lot's of compiler warnings from the generated api classes like this one:
`The corresponding parameter in the supertype 'Popup' is named 'event'. This may cause problems when calling this function with named arguments.`
Which comes from: `override fun _gui_input(arg0: InputEvent)`

But this is mainly an issue of the godot api json as the arguments are named like this there.
A workaround could be to get the argument name from the supertype in the json and not from the definition itself. Or we could open an issue on the gdnative repository directly so they fix the root issue.

Also we still use deprecated code in the `godot-library-extensions` regarding coroutines.

We can then concentrate on the real warnings and users may not run into trouble because we generated suspicious code.

Depends on the solution we take.

Access the supertypes methods argument names if possible, fallback to provided argument types in json.

We could suppress the warnings which is not the way to go.

Our code generates those warnings during compilation.

lassem:
Fixes #16 

Fixes compile warnings by trying to be clever when reading api.json. 
Sanitise method argument names.

When resolving the argument name, traverse the tree to find the parent method (if any) and use the argument name from there. As suggested by @chippmann in #16


piiertho (approved):
Thanks for contribution ! Nothing to say on code.

Issue #10, opened 2020-02-06
Add JVM/JNI target for local development


JustinMullin:
**Describe the problem or limitation you are having in your project:**
The Kotlin/Native compiler is very slow at present, resulting in fairly long turnaround for small game code changes using godot-kotlin during development.

**Describe how this feature / enhancement will help you overcome this problem or limitation:**
For local development, running game logic on the JVM rather than native code would allow for quick re-compile times, as well as some JVM niceties such as live class reloading at runtime and direct debugging+breakpoints into the Kotlin code. The Kotlin/Native implementation in godot-kotlin can then be used in the exported game to remove the dependency on the JRE.

**Describe implementation detail for your proposal (in code), if possible:**
- Core+generated classes in godot-kotlin are extracted as Kotlin Multiplatform `expect` declarations. Native implementations of these classes will acquire `actual` keywords to define them as platform implementations, but otherwise remain unchanged.
- Kotlin/JVM `actual` implementations are created utilizing `external` methods to call out to Godot APIs; implementations will be registered by the JNI later.
- Game logic uses the same code as before, but for local development is compiled using the standard Kotlin/JVM compiler and the JVM `actual` implementations rather than the Kotlin/Native compiler against the existing implementations. The output classes are bundled to a JAR and output into the Godot project path.
- A GDNative library is created which links to the JDK. This library is relatively project agnostic and shouldn't need to be recompiled during regular development. Upon gdnative_init, it instantiates a JVM via the JNI API, including the game logic JAR on the classpath. This initialization process also registers implementations for the JNI bindings above, allowing the JVM code to interface indirectly with GDNative. The implementations would be largely pass-through, but need to handle a bit of memory management and translation as the signatures for the `external` methods don't support pointer types, etc.

**Is there a reason why this should be in this project and not individually solved?:**
Ideally a JVM+JNI implementation will be maintained in parallel to the Kotlin/Native version to ensure seamless transition between fast local development and export.

JustinMullin:
Sorry if this is a bit hand-wavey on implementation details. I was able to get some basic game logic running on the JVM and interfacing bidirectionally with Godot with the design described above. Working on cleaning up a local POC of this that I had built against MrAkakuy's project. It had some rough edges and is now pretty out of date with the updates you folks have been making (more so if the refactorings to use Poet go through), so not the most usable thing in current state.

JustinMullin:
[...]

PR #20, opened 2020-02-10
Extract expect/actual structure for core+generated types

JustinMullin:
Extract base structure for wrapper library to a common module using `expect` declarations, and refactor native implementations to use `actual`. Move shared methods on core types to extension functions to reduce duplication between platform-specific implementations.


JustinMullin:
Opening this to preview what utilizing the MPP `expect`/`actual` keywords would look like for this project, as this is likely to be a sticking point in implementing https://github.com/utopia-rise/godot-kotlin/issues/10. There's some overhead introduced here, and a couple of design considerations.

If it's decided that godot-kotlin is moving this direction, it would be good to get the refactoring out of the way, as it does involve some pretty significant shifting of code and this branch will drift rapidly as modifications are made to core classes. I also recognize that the JVM platform implementation is a lower priority, so don't want to push too hard on this path if there are concerns.

A few highlights/potential places where this may be controversial:
- The `expect`ed class can't contain default implementations, so Godot-external logic in the core classes needs to move to extensions on the core class (the route demonstrated in this PR), or otherwise be duplicated in each platform-specific implementation.
- With the extensions route, there are some ramifications for imports, as each extension method is an individual import. As a result, many imports may be needed in some cases, for users that are disinclined to use the wildcard import `import godot.core.*`
- An `expect` class can't define `const` `val`s. I'm not sure if this is a restriction that will be lifted at a later date by K/N, but it's true at present. Here I've removed the `const` modifier from `val`s that manifest in the core/generated classes; I don't believe this should cause any problems as inlining wasn't likely gaining much in these cases, but it could be a future consideration.

piiertho:
Hi ! 
Great work !
I cannot build godot-library with parameter platform, using:
`./gradlew :wrapper:godot-library:build -Pplatform=macos`
(It does metadata but no compilation)

JustinMullin:
Hey @piiertho - do you mean you're seeing the `godot-library-metadata` artifact published but not the `godot-library-macos` artifact? I'm able to build the `godot-library-macos` artifacts using that command locally; cleaned the Gradle project and cleared Maven local artifacts, and was able to build+run the sample games from scratch. The CI check seems to be successfully running `:wrapper:godot-library:compileKotlinMacos` too here. It's entirely possible I'm missing something cached somewhere locally though.

piiertho:
Oh sorry my bad, was trying to compile Windows from the wsl commandline

CedNaru:
Impressive Job.

Is there some overhead when using extension fonctions instead of regular methods ? 
Otherwise I see no issue with using them. There is no private properties needed in the common code so I don't think we will encounter any visibily issue

JustinMullin:
Nope, no overhead to extension functions (other than potentially a minimal compiler cost). They end up as static methods with the receiver as a parameter in the bytecode, so just a regular method invocation.

samples/games/kotlin/gradle.properties
```diff
@@ -0,0 +1 @@
+platform=macos
```
piiertho: seems to have commit a wrong file :)

piiertho:
@JustinMullin you put this PR as a draft, do you want to continue working on it or should we review it ?

samples/games/kotlin/gradle.properties
JustinMullin: Whoops! Deleted now, good catch.

JustinMullin:
> @JustinMullin you put this PR as a draft, do you want to continue working on it or should we review it ?

I opened as draft expecting this might initially just generate conversation vs. be immediately merged, but it seems it's generally pretty accepted as an approach here, so I'll mark ready for review to kick that off formally. =)

chippmann:
@JustinMullin Really great work! Thanks a lot for your contribution!
@piiertho I wouldn't merge it though. I would first merge the entry generation branch once it's finished, so we can make a release first. After that, we can merge this i think as after that this will be the biggest task.
Are you all ok with that?

piiertho:
@chippmann I'm ok with that, I will complete Android and iOS compatibility too, before making a release.

chippmann:
@JustinMullin If you are ok with it I'll close this PR as it becomes obsolete and incomaptible with the merge of the `master-merge` branch. As said we can discuss the whole JVM and expect/actual topic once the new binding is in place :-)

chippmann:
Closing in favor of `master-merge`

PR #26, opened 2020-02-13
Feature/project layout

Wavesonics:
This is not ready to merge, but I wanted some feedback.

In the `game` sample directory, I moved the `project` contents up one level. So the `kotlin` directory is now a child of the Godot project.

As a child, the `kotlin/bin` dir can now be referenced from the Godot projects, so the GDNative resource can point to the binaries via `res://kotlin/bin/xxx`

There's still kind of a lot of faff in the kotlin dir so I think keeping it some what separate from the Godot files is nice. Plus we can easily tell the Godot project to just ignore the kotlin dir.


chippmann:
@Wavesonics you are aware of the fact that you also need to change the paths inside the `build.gradle.kts`, and the `samples.gdnlib` right?
Apart from that, i don't have to say much about it. I'm fine with it i guess.
Maybe we want to change the name of the kotlin dir?

piiertho:
I think we can go for something like src/kotlin or script/kotlin.
Also it could be cool to set compiled code in platform directories like bin/macos/libkotlin.dylib or bin/android/arm64/libkotlin.so. So that we do not overwrite other os librairies.

Wavesonics:
@chippmann yeah, I just wanted to get people's input before I started going any further down this road.

Another option: we could stick with something similar to the original setup, where the Kotlin is totally separate, and just have the build script copy the binaries into the godot project. How do the C# projects handle this?

Wavesonics:
@piiertho I thought about src, but the kotlin dir also has an src under it, so: `src/kotlin/src` seemed like it could be confusing. However `script` might be nice. I imagine a truly mad world where script is a common dir used by all languages, and you could mix C# (script/csharp), Rust (script/rust), and Kotlin all in the same project xD

piiertho:
Personnaly I'm fine with this layout.

lassem:
We also have `.gdignore` if there are things we want to hide from the ide.

Wavesonics:
Ok I think the latest CL is in a pretty good spot. Let me know if there is anything else that needs changing.

piiertho (approved):
This layout seems definitely better than older :)
As it is important changes (it impacts end user experience), I'd like to have @chippmann  and @CedNaru approval too.

CedNaru (approved):
I'm all for having the kotlin project inside the godot one.
So I'm okay with the changes.

samples/coroutines/coroutines.gdnlib
```diff
@@ -7,12 +7,10 @@ reloadable=true
 
 [entry]
 
-OSX.64="res://libkotlin.dylib"
-Windows.64="res://kotlin.dll"
-X11.64="res://libkotlin.so"
+Windows.64="res://kotlin/build/bin/windows/releaseShared/kotlin.dll"
```
chippmann: On the games sample project we link the debug version and here the release version. I think we shouldlink the debug version both times as they are samples not "finished" games and its more consistent.

chippmann (dismissed):
Apart from that one change, i'm fine with it. Thanks a lot for your work :-)
@piiertho you can merge it once this branch is rebased onto the changes made in the android branch :-)

samples/coroutines/coroutines.gdnlib
piiertho: @Wavesonics, can you rebase on master and add requested changes so that we get that merged :)

piiertho: Oh did not see it's coroutine module, it is compiled as release in gradle build script. Can you set it in debug for the moment ? We will add a target parameter to define release or debug.

piiertho (changes requested):
Needs rebase on master.

Wavesonics:
Yeah I can do this tonight

Wavesonics:
@piiertho ok I made both samples build debug, and I rebased on master.

Should be good to go now.

piiertho (approved):
Thanks !

piiertho:
@chippmann As the changes you requested were done, I consider that ti is ok :)

chippmann:
> @chippmann As the changes you requested were done, I consider that ti is ok :)

@piiertho Sure! Thank you :-)

PR #32, opened 2020-02-18
hotfix/add set to dictionary

piiertho:
Add to Dictionary operator fun get and set, add auto cast method from any to variant to make it easier to use.

This enable user to write code like:

```kotlin
dict["test"] = 1001
dict[Vector2(0, 1)] = Vector3(0, 1, 2)
dict.erase("test")
```

We should wait that @2shady4y tests it on his sqlite implem.

### DIFF
--- a/wrapper/godot-library/src/main/kotlin/godot/core/Dictionary.kt
+++ b/wrapper/godot-library/src/main/kotlin/godot/core/Dictionary.kt
@@ -53,16 +53,25 @@ class Dictionary: CoreType {
 
     fun erase(key: Variant) = godot_dictionary_erase(nativeValue, key.nativeValue)
 
+    fun erase(key: Any) = this.erase(Variant.from(key))
+
     fun has(key: Variant): Boolean = godot_dictionary_has(nativeValue, key.nativeValue)
 
-    fun get(key: Variant): Variant? = godot_dictionary_operator_index(nativeValue, key.nativeValue)?.pointed?.readValue()?.let { Variant(it) }
+    fun has(key: Any): Boolean = this.has(Variant.from(key))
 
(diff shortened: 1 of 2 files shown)


wrapper/godot-library/src/main/kotlin/godot/core/Dictionary.kt
```diff
@@ -53,16 +53,25 @@ class Dictionary: CoreType {
...
-    fun get(key: Variant): Variant? = godot_dictionary_operator_index(nativeValue, key.nativeValue)?.pointed?.readValue()?.let { Variant(it) }
+    fun has(key: Any): Boolean = this.has(Variant.from(key))
+
+    operator fun get(key: Variant): Variant? = godot_dictionary_operator_index(nativeValue, key.nativeValue)?.pointed?.readValue()?.let { Variant(it) }
+
+    operator fun get(key: Any): Variant? = this[Variant.from(key)]
 
+    operator fun set(key: Variant, value: Variant) = godot_dictionary_set(nativeValue, key.nativeValue, value.nativeValue)
+
+    operator fun set(key: Any, value: Any) = this.set(Variant.from(key), Variant.from(value))
```
CedNaru, line 70: Wouldn't this be better to use CoreType instead of Any ?

piiertho: No, because we had to cast kotlin's primitive too
image

CedNaru (approved):

2shady4u:
Heyo, I'll test this as soon as possible (too much work atm)

Would it also be possible to add get with a default value?
Like this:
Capture

piiertho:
I added it right now, but did not test it, I Can test it this evening too. If it works on your side I'll merge it

CedNaru (approved):

2shady4u:
So yeah, it works as intended for me 😄 
I'm having some problems converting kotlin cinterop variables to Godot Variants, but that's unrelated to this PR.

Capture

PR #33, opened 2020-02-20
Godot Editor plugin

Wavesonics:
~~This is an extremely rough proof of concept.~~ _This is looking pretty good now I think. Still more work to do, but it's in a pretty workable state as is._

After investigating IntelliJ project templates, and IntelliJ new project wizards, each of them had significant drawbacks.

So here is my 3rd attempt: A more Godot-centric approach.

The user flow in real life would be something like this:

1. Create a new totally standard project in Godot
2. Download the Kotlin integration plugin from the asset library
3. Enable the plugin (I always forget this step lol)
4. Open the Kotlin tools dialog from Project -> Tools menu
5. Click "Setup"

*The plugin takes over from here:*

1. Download the kotlin project template (_the contents of the /kotlin dir in our current setup_)
2. Extract the template zip
3. Delete the template zip
4. ~~Edit the gradle file with project name and such (_not implemented yet_)~~
5. Call gradle to configure the whole shabang
6. Create the gdnlib file ~~(_not implemented yet_)~~

Now that everything is setup, we provide a toolbar icon to trigger the a gradle build. ~~(_not implemented yet_)~~

Limitations right now:

- I could not for the LIFE of me get `OS.execute()` to execute gradlew with an argument. On it's own it's fine, but trying to pass `build` to it failed every why I could think of. So as a work around I included a batch file that calls gradle w\ the `build` argument. ¯\_(ツ)_/¯ If we have to ship with this I think it's still fine, but I'd love to figure out why I can't pass in the damn argument normally...
- This proof of concept only works on windows, but shouldn't take long to add *nix support.


Wavesonics:
Ok, things are actually starting to look a little better. I have the core functionality working reliably (_as far as I can tell_) and looking a little better.

I stubbed out Unix support, but I need someone on OSX or Linux to try it and help tweak the shell scripts.

The one thing that isn't really handled is the `config()` part of the gradle script, which currently needs to be entered manually for each `classes.json` file. However, I think maybe we should just have the script scrape the directories and discover those files for it's self, removing one more maintenance step.

Wavesonics:
Oh I must have been confused, there was a conversation that made me think we were running out of options on the annotation processor, and it might end up not working.

Also even if it did work I thought the annotation processor would just be producing these json files and we'd still need to point the build script at them somehow, how's it planned to work so that we don't need the json at all?

Well anyway that's all great news!

GaidamakUA:
@Wavesonics I'll try it on MacOS first thing tomorrow.

Wavesonics:
Awesome thanks @GaidamakUA 

The gradle interactions almost certainly won't work right out of the box. But there are only two places that I'd need your help to tweak:
`godot-kotlin\tools\godot-editor-plugin\addons\kotlin\tools\GradleUtilities.gd`
```
static func execute(command: String):
	var output := []
	
	if is_windows():
		OS.execute("CMD.exe", ["/C cd kotlin && %s" % command], true, output)
	else:
		OS.execute("sh", ["cd kotlin && %s" % command], true, output) # <- This line
	
	print(output)
```

And this file `godot-kotlin\tools\kotlin-project-template\build`
```
#!/usr/bin/env sh
./gradlew build
```
Actually that last file should be fine, but if not, that's one of the few OS specific areas.

GaidamakUA:
@Wavesonics `kotlin/build` doesn't work due to Windows style line endings. Please, configure your editor to use unix style line endings since they work with Windows anyway and resave.
Description of the issue: https://stackoverflow.com/questions/29045140/env-bash-r-no-such-file-or-directory
The same thing needs to be done with `kotlin/gradlew`.
Also it's necessary to give execution permission to `kotlin/build` and `kotlin/gradlew`. Something like `chmod +x ./gradlew`

piiertho:
Here is what I get on Mac OS:

image

Got problem for loading kotlin Icon and update spinner on top right of screen does not stop.
but it seems to generate gradle files, great work !

Wavesonics:
@piiertho I think I've fixed the icon loading issue.
@GaidamakUA huh my git client just keeps discarding the changes when they are only line endings. I'll have to take a look at this later tonight.

Wavesonics:
**ToDo:**
- ~~Fix line endings (_Unix style_)~~
- ~~Add execute permission to *nix scripts~~
- ~~Get `OS.execute()` working on Linux~~

Wavesonics:
Ok I have things working on Linux now. But I don't have an OSX machine to test on, so if someone could pull this and give it a try that'd be great. @piiertho 

Annoyingly, on Linux at least, if the `OS.execute()` command fails, you get nothing in the output, which makes debugging it quite hard. If you see `[]` in Godot's output, that means the command that was passed to `OS.execute()` failed. There should be a number above the brackets in output that is all we'll have to go by, it's the exit status of the command.

piiertho:
It seems to successfully start gradle script. But it fails when generating entry file.
image

EDIT:

Here is log from command line.
```
./gradlew build

> Configure project :
Kotlin Multiplatform Projects are an experimental feature.
Configuring target linux
Configuring target macos
Configuring target windows
Configuring target androidArm64
Configuring target androidX64
Configuring target iosArm64
Configuring target iosX64

Some Kotlin/Native targets cannot be built on this macos_x64 machine and are disabled:
    * In project ':':
        * target 'windows' (can be built with a mingw_x64 host)
To hide this message, add 'kotlin.native.ignoreDisabledTargets=true' to the Gradle properties.


> Task :androidArm64MainGenerateEntry FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':androidArm64MainGenerateEntry' (type 'DefaultTask').
> Cannot write to directory '/Users/piertho/github/godot-kotlin/tools/godot-editor-plugin/kotlin/build/godot/entries/androidArm64Main' specified for property '$1', as ancestor '/Users/piertho/github/godot-kotlin/tools/godot-editor-plugin/kotlin/build' is not a directory.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.0.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 1s
1 actionable task: 1 executed
```

Also it will compile for all platforms by default. Maybe beeing able to choose a platform, an arm architecture if necessary and the build type (debug or release), could be a great idea.

Wavesonics:
Status of this PR:
- OSX and Linux now work
- debug/release can be selected in the UI

What's left:
- Waiting on @chippmann entity generation branch. Once it's in the template script can remove the `config()` section, which since it is manual, currently breaks these builds as the plugin doesn't update it.
- ~~I'm going to add a platform selector, so in the UI you can decide which platforms get built~~ **complete**
- Create a gradle task to zip up the project template `tools\kotlin-project-template` and publish it as an artifact to a github release. Once that is done, we can update the download URL in the script to point to `releases/latest/kotlin.zip`

Wavesonics:
_We discussed this on discord, but putting it here for completeness:_
The failure @piiertho mentioned above is due to the json entities. The template build script will be updated to remove the manual `config()` section once the new entity generation code lands. And at that point this should be able to fully build a project automatically.

Other that the `config()` issue, I think with my latest commit that this is actually good to go for a first release.

Wavesonics:
Once we merge #42 I'll get the gradle script in the template updated, and then this should be good.

chippmann:
> Once we merge #42 I'll get the gradle script in the template updated, and then this should be good.

Very cool! I'll look into this now as well as i now finally have some time ^^ Looks incredibly promising.

Wavesonics:
Ok I rebased on the new entity_generation branch, and made some final modifications for the plugin to work with it. It should successfully setup and build a new project at this point.

A couple notes:
- I added an ExampleNode class so new users can see the default place to put their classes
- I added a gradle task to zip up the kotlin project template. It deposits it in `<root>/build/kotlin_template.zip` this is the thing that needs to be put some where for people to download. Currently it is uploaded to my personal server. Before we release this we need to decide where that zip will be hosted, and then change the URL in the plugin. Ideally, the gradle task that creates the zip could publish it to where ever it is supposed to be.

I will be out of contact for the next 2 days, if anyone wants to make the final changes and merge this in thats cool. Also if others can test it out and make sure it works on macOS and Linux that'd be great too.

CedNaru:
Can't we just create an individual project  template repo  in Github and make the plug-in download it directly ?

Wavesonics:
OK this is ready for review now.

There's certainly more that can be added to it, but this is a solid first version that I think makes setting up a new project a pretty nice experience for new users.

I changed the URL to point to this repo, so it won't work until we have at least one release containing the project template, as such:
https://github.com/Wavesonics/godot-kotlin/releases/tag/PluginTest

Maybe just for testing purposes we could create a similar release here.

Wavesonics:
@CedNaru ~so the project does download the template directly from GitHub. I was thinking we'd host a zipped version of the template in the releases section of this repo, but it's very easy to change that of course.~

~I personally don't see much advantage to putting it in a second repository, but maybe someone else has a good reason?~

You probably mean downloading the whole repo like this:
https://github.com/Wavesonics/godot-kotlin/archive/master.zip

Thats an idea, though we can't as easily control versioning of it. Maybe we need a change that is only compatible with the new release of the plugin, ect.

Wavesonics:
Ya know, the more I think about it @CedNaru that's kinda a cool idea.

What does everyone else think?

chippmann:
I like the idea as well.
Regarding versioning. We could tag them and download the state tagged as zip.
Only problem i see: we need to find a way to do this automatically. Otherwise i think sometime in the future it will happen that we publish a release, but forgot to tag the corresponding version in the templete repo.

CedNaru:
Maybe we can close this PR as the plugin got its own repo now.
https://github.com/utopia-rise/godot-kotlin-editor-plugin

chippmann:
I think so as well. I close it. Feel free to reopen it if you think otherwise.

Issue #28, opened 2020-02-15
Add choice for target between RELEASE and DEBUG


piiertho:
**Describe the problem or limitation you are having in your project:**

For now, release or debug is hardcoded in gradle build script, we should be able to specify a release target or to have one by default, that is the same for all projects (coroutines project is in RELEASE mode).

**Describe how this feature / enhancement will help you overcome this problem or limitation:**

User would be able to specify a release target.

**Show a mock up screenshots/video or a flow diagram explaining how your proposal will work:**

None to provide

**Describe implementation detail for your proposal (in code), if possible:**

1) Add a target parameter, which would be called with `-Ptarget=RELEASE/DEBUG`:
```kotlin
val target: String by project
```
2) Modify this block to conditionally select target:
```kotlin
this.target.binaries {
    sharedLib(listOf(org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType.RELEASE))
}
```

**If this enhancement will not be used often, can it be worked around with a few lines of code?:**

It depends on build scripts.

**Is there a reason why this should be in this project and not individually solved?:**

We should be able to publish release

Wavesonics:
Recommend closing this issue as PR #35 implemented this.

chippmann:
> Recommend closing this issue as PR #35 implemented this.

True. Closing it.

PR #35, opened 2020-02-21
Added selectable build target using -PbuildTarget=release

Wavesonics:
Implementation for #28
Just `target` was a loaded term within the build script, so I opted for `buildTarget` to differentiate it.

In my other PR #33 I establish a template for the kotlin side of projects `tools/kotlin_project_template`. Once we pull that in we can codify changes like this in the `build.gradle.kts` file in there and use that as the standard.

### DIFF
--- a/samples/coroutines/kotlin/build.gradle.kts
+++ b/samples/coroutines/kotlin/build.gradle.kts
@@ -1,7 +1,7 @@
 val platform: String by project
 val armArch: String by project
 val iosSigningIdentity: String by project
-val buildTarget: String? by project
+val buildType: String? by project
 
 buildscript {
     repositories {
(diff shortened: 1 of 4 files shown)


samples/coroutines/kotlin/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 val platform: String by project
 val armArch: String by project
 val iosSigningIdentity: String by project
+val buildTarget: String? by project
```
chippmann: Actually I'm also against `buildTarget`. A target implies for me a targeted platform in a multiplatform project (which we are).
So how about `buildType` like it's called on the kotlin side anyways?

samples/games/kotlin/build.gradle.kts
```diff
@@ -115,7 +116,15 @@ kotlin {
             if (this is org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeCompilation) {
                 println("Configuring target ${this.target.name}")
                 this.target.binaries {
-                    sharedLib(listOf(org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType.DEBUG))
+                    val libTarget = when(buildTarget?.toLowerCase()) {
+                        "release" -> listOf(org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType.RELEASE)
+                        "debug" -> listOf(org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType.DEBUG)
+                        else -> {
+                            System.err.println("Build target not specified, defaulting to DEBUG. To set release target, specify: -Ptarget=RELEASE")
```
chippmann: A warning would be sufficient here. Personally i only log errors when they are fatal or non fatal but critical. This is neither of those things and warning is also red 😜

chippmann (changes requested):
Some minor details (mostly personal preference) but very cool! Thanks!

samples/coroutines/kotlin/build.gradle.kts
piiertho: I think we should discuss of that, in godot-cpp it is target, which I don't like too...

piiertho (changes requested):
Can you add the same code in the documentation ? It would be awesome :)

samples/coroutines/kotlin/build.gradle.kts
Wavesonics: Ya `buildType` sounds good to me

Wavesonics:
Ok, added a line to the documentation about how to use this, and changed it to `buildType`

chippmann (approved):
From my side ok. Thanks for the changes.

piiertho (changes requested):
Only variable has been added to getting started's final build script, it's missing the use of it

piiertho (approved):
Thank you for the contribution !

PR #40, opened 2020-02-23
Hotfix/fix travis ci deploy

piiertho:
This fixes artifacts deployment and resolution on bintray. It enables user to use our bintray to get artifacts, as FakeRelease proves it.
Here is bintray with publish test: https://bintray.com/beta/#/utopia-rise/kotlin-godot
Also I will delete FakeRelease and artifacts on bintray when merged.


chippmann (approved):
I tested it on linux and everything works now as it should.

Thank you very much for doing this cumbersome CI work! Really appreciate it.

PR #41, opened 2020-02-23
Fix argument name in compile from source doc

Wavesonics:


chippmann (approved):
Thanks a lot for the fix!

Issue #36, opened 2020-02-22
Coroutines crash on startup.


piiertho:
**Version:**
Master branch

**OS/device including version:**

MacOS Catalina 10.15.3

**Issue description:**

Coroutines sample crashes on macos when starting, with stacktrace:
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff70c3d7fa __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff70cfabc1 pthread_kill + 432
2   libsystem_c.dylib             	0x00007fff70bc4a1c abort + 120
3   libkotlin.dylib               	0x0000000121e46de9 konan::abort() + 9
4   libkotlin.dylib               	0x0000000121e4adae TerminateWithUnhandledException + 14
5   libkotlin.dylib               	0x0000000121e4acf1 KonanTerminateHandler() + 193
6   libc++abi.dylib               	0x00007fff6dc6bdc7 std::__terminate(void (*)()) + 8
7   libc++abi.dylib               	0x00007fff6dc6bb6c __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 27
8   libc++abi.dylib               	0x00007fff6dc5d45d __cxa_throw + 113
9   libkotlin.dylib               	0x0000000121e62afc ThrowException + 124
10  libkotlin.dylib               	0x0000000121c70f44 kfun:godot.utils.getMB$godot-library(kotlin.String;kotlin.String)ValueType + 1092 (Utils.kt:15)
11  libkotlin.dylib               	0x0000000121abf94d kfun:godot.Object.$INITIALIZER$lambda-38#internal + 45 (Object.kt:154)
12  libkotlin.dylib               	0x0000000121ac3595 kfun:godot.Object.$$INITIALIZER$lambda-38$FUNCTION_REFERENCE$1438.invoke#internal + 37 (Object.kt:153)
13  libkotlin.dylib               	0x0000000121ac364f kfun:godot.Object.$$INITIALIZER$lambda-38$FUNCTION_REFERENCE$1438.$<bridge-UNN>invoke()ValueType#internal + 143 (Object.kt:153)
14  libkotlin.dylib               	0x000000012186c129 kfun:kotlin.native.concurrent.FreezeAwareLazyImpl.<get-value>()#GENERIC + 2793 (Lazy.kt:52)
[...]

PR #42, opened 2020-02-24
Implement annotation based registration

### LINKED ISSUE #2: Use kotlin compiler plugin and reflection to generate code [Moved from old repo]
Currently the `api-classes-generator` and the `entry-generator` use a errorprone and hard to understand code generation using stringBuilders.
The code would be much cleaner and easier to maintain if we use reflection to generate the api-wrapper and annotations for the entry-classes generation.
Also it would make the project easier to use as a user doesn't have to write config json's anymore.

One no longer needs to configure the errorprone json files and maitanance is much simpler.

None to provide.

Use the compiler plugin created in issue #13 for the config generation.
~~Maybe use [kotlinPoet](https://github.com/square/kotlinpoet) for the api classes generation~~ -> done in #14 

Yes with the current system. But it's not user friendly at all.

Yes as the code generation is the main part of the project.

### LINKED ISSUE #27: public lateinit Vector2 is unassigned if has a value of 0,0 in Godot editor
**Version:**
Godot 3.2

**OS/device including version:*
Windows 10

**Issue description:**
If your Kotlin class has a public lateinit var of Vector2, it's "unassigned" value in the Godot editor is 0,0. Giving it any other value will cause it to be assigned in Kotlin, and everything is fine. However, the user may want 0,0 as their value, and in this case, accessing the var in Kotlin will result in an unassigned lateinit var exception being thrown.

**Steps to reproduce:**
Run the coroutines sample project.
`MirrorBall.kt`
`lateinit var center: Vector2`

**Minimal reproduction project:**
The coroutines sample project.

chippmann:
# Overview
This implements annotation based Registration of Classes, Functions, Properties and Signals.

Information on how to use the annotations are provided in the Registration readme.

# Technical background:
The annotation processing is done using a KotlinCompiler plugin I implemented in the old repo (`kotlin-compiler-plugin` and `kotlin-compiler-native-plugin` modules). To read the annotations on native targets we use the [MpApt project](https://github.com/Foso/MpApt). The annotations are received in a AnnotationProcessor (the `godot-annotation-processor` module).
The Entry file is then generated using the `EntryGenerator` (inside the `entry-generator` module).

The problem we had is that the compiler plugins are called too late in the compilation process for the compiler to pick up the newly generated `Entry.kt` file.
Jetbrains goes around this problem by directly generating JVM Bytecode / IR (for native) and feeding that into the compiler (see the [Serializable compiler plugin](https://github.com/JetBrains/kotlin/tree/master/plugins/kotlin-serialization)).
As I don't have the required know how to directly generate IR (nor do i find it useful as we cannot debug it conventionally) I opted to overcome this problem by adding a dummy target during build. As the actual compilation task is the shortest one the time overhead is negligible IMHO. During the compilation for this dummy target, the `Entry.kt` file is generated. So it's already present when we compile the real target. This dummy target logic is located in the `KotlinTargetPreset` (in the `godot-gradle-plugin` module). It ensures also only one dummy target is added even when building for multiple targets.

# Summary of the steps:
1. During build a dummy target is added
2. The dummy trarget get's compiled
3. The kotlin compiler plugin calls the annotation processor with all required annotations
4. The annotation processor gathers all information required for the entry generator
5. The annotation processor calls the entry generator with all required informations
6. The entry generator generates the entry file
7. The actual target is compiled including the now present entry file
8. rest of the normal build pipeline follows...

# Alternatives that i tried:
- Normal annotation processing using `kapt`:
The problem here is, that `kapt` only works for JVM targets. There is no `kapt` for native targets. Once #10 is implemented. My goal is to switch to normal `kapt` annotation processing as it's much faster. The migration should be almost seamless as i already use a annotation processor and we just need to remove the compiler plugin part and the dummy target.
But for now this is sadly not possible.

- Extracting the kotlin compiler frontend into a script and call it directly:
This worked for the most part. The problem was that we needed to add all required sources of all dependencies to be parsed as well. Which is no problem for our own code but is a problem for dependencies the user might add. Furthermore the parsing took a long time (around 45-60s on a Dell XPS 15  with an i7-8750H). It also takes this long for the `godot-library` so it was not a problem with the script, more of the sheer amount of code that needed to be parsed.
This and the fact that we cannot gather all the required source code for all the dependencies a user might add, lead me to the conclusion that i don't want to use this approach.

# The next steps:
There are many things we can optimize regarding annotation processing. But I didn't implement it yet as this is the basics on which we can build on.
Some improvements i want to make:
- Implementing some extension functions to make the life easier when working with signals
So a user could call `connect(::methodInThisClass)` and this would automatically connect the signal he has overriden
- Use code generation to provide the possibility to export `NodePath` from annotating a property of a type which derives from `Node`.
Like this a user could write this code:
```kotlin
@Export
lateinit var body: KinematicBody
```
instead of:
```kotlin
@RegisterProperty(true, null)
lateinit var body: KinematicBody

@RegisterFunction
override fun _ready() {
    body = KinematicBody2D from getNode(NodePath("Path/To/Node"))
}
```
and then could wire up the node path in the editor, like when writing 
```
export var pathToBody : NodePath
onready var body: KinematicBody = get_node(pathToBody)
```
 in GDScript.

# Notes
**Note:** I did not bother migrating the coroutines sample as it does not work at the moment anyways and we can do the migration together with the fix of it in #36.

@piiertho please check the bintray configuration with care as I'm not sure whether they are correct.
@CedNaru Please check the registration readme with care to be sure everything is clear (I'm sadly not that good at writing documentation in an understandable manner)

I tested this on linux and android. Please also test on MacOS and Windows.

This resolves #2 and closes #27 .


Wavesonics:
I've got fairly poor service here and can't seem to comment on the exact line in the change, but the new config section in the `build.gradle.kts` scripts needs to take the `buildType` project var into account. Currently it just hard codes to debug.

chippmann:
> I've got fairly poor service here and can't seem to comment on the exact line in the change, but the new config section in the `build.gradle.kts` scripts needs to take the `buildType` project var into account. Currently it just hard codes to debug.

True i forgot that. It's added now. Thanks :-)

piiertho (changes requested):
Hi ! Really great job !
There is some mission documentation in `COMPILING_FROM_SOURCES.md`, to compile `annotations-internal` only.
Everything working on MacOS.

chippmann:
> Hi ! Really great job !
> There is some mission documentation in `COMPILING_FROM_SOURCES.md`, to compile `annotations-internal` only.
> Everything working on MacOS.

Whops. Sorry about that. Added it now.

chippmann:
I rebased this PR to include the changes made in #41

piiertho (approved):
Thanks a lot for this awesome work !
It's working well on MacOS and Windows on my side :)

chippmann:
> Thanks a lot for this awesome work !
> It's working well on MacOS and Windows on my side :)

Good to hear! Thanks for your review.

CedNaru (approved):
Really great job.
I'm don't know much about Kotlin/Gradle plugin and Poet so I'm a bit lost in a big part of the code.
But the final product is really great to use. There is obviously pists to improve the semantic/syntax for the user. But as a first version, it's already good enough.

PR #44, opened 2020-02-25
Set source target compatibility to java 8

chippmann:
This should prevent the source code to be build for higher java versions when built locally, so it stays compatible with whatever java version the user is using to build he's games.

### DIFF
--- a/tools/annotations/build.gradle.kts
+++ b/tools/annotations/build.gradle.kts
@@ -20,6 +20,11 @@ val bintrayKey: String by project
 val platform: String by project
 val armArch: String by project
 
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
 kotlin {
     if (project.hasProperty("platform")) {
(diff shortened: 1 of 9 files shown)


CedNaru:
Simple change. Good for me.
But if the user still import the project in IDEA with the option to use a custom version of Java instead of the one defined in Gradle, building from Idea will fail like before right ?
We might need to write a little disclaimer in the readme.

chippmann:
> Simple change. Good for me.
> But if the user still import the project in IDEA with the option to use a custom version of Java instead of the one defined in Gradle, building from Idea will fail like before right ?
> We might need to write a little disclaimer in the readme.

Maybe I'm not yet sure. I'm trying this first.

chippmann:
@CedNaru No it seems to be fine. I set the jdk version to 13 in the IDE for all modules and the samples to 1.8 and it builds fine.

chippmann:
Even though it is a small change, I want to have @piiertho 's review as well as he has some good knowledge as well regarding gradle.

piiertho (approved):
It's ok for me :)

PR #49, opened 2020-02-28
Fix entry generation tasks for projects with additional configurations

chippmann:
For the entry generation we add an additional compilation step to generate the entry files (a dummyTarget).
With the previous approach it worked as long as the user had no other configurations on his targets than normal dependencies.
But if the user defines other configurations like cinterops for his game (like @2shady4u does in his sqlite-wrapper) the entry generation compilation task failed as no cinterop was done for it as it was not configured for it.

This PR let's the user configure his targets in a separate function that is passed to our plugin and is added to each target the user configures including our dummyTarget.
Like this all configurations the user applies to his targets, are also applied for the dummy target.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -127,9 +127,9 @@ kotlin {
 ```
 
 \
-As the last step, we need to specify for which target to build for:
+Then, we need to specify for which target to build for:
 ```kotlin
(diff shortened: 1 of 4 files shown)


piiertho (approved):
Tested on OSX, it works well. 
Thanks for this clever fix, really love it :)

PR #58, opened 2020-03-04
Fix deploy annotation-internal and add jvm deployment

piiertho:
0.1.0-3.2 release went bad :/
Here is a PR to make release great again :P

This adds parameter to specify jvm publication.
This fixes annotations-internal bintray conf.
This adds jvm deploy to travis.


chippmann (approved):
Looks good to me.
Thanks for the fix!

PR #59, opened 2020-03-04
Feature/game of life benchmark

CedNaru:
GameOfLife benchmark is ready.
Build, open the godot project, inspect the main node and change the language you want to use. 
Launch the game and enjoy how quick Koturtle can be.

PR #60, opened 2020-03-04
Feature/bunny benchmark

CedNaru:
BunnyMarK benchmark is ready.
Build, open the godot project, inspect the main node and change the language you want to use.
You can choose between 3 different implementations of the benchmark as well.
Launch the game and enjoy how quick Koturtle can be.


raniejade:
Out of curiosity, can you post the numbers here? 🙏

CedNaru:
Running on I5 4690K, 16GB, Windows 10:
TextureV1:
Kotlin -> 1100 bunnies
GdScript -> 11000 bunnies

SpritesV1:
Kotlin--> 1564
GdScript --> 9015

V2:
Kotlin --> 35
Gdscript --> 9600

Number of bunnies before reaching 60 fps, higher is better.

raniejade:
Ohh, wow! I'm going to try this on my binding (It's probably much slower since I didn't re-implement the core types - and just delegate every operation to Godot).

CedNaru:
If only it was the calls to Godot that were slow. But actually what takes the most time is to instanciate the Sprite Nodes each frame(benchmark V2). Nothing much going on there, we just use the pointer returned by Godot and wrap it with a Kotlin Sprite object.

samples/benchmark/bunnymark/Test.tscn
```diff
@@ -0,0 +1,10 @@
+[gd_scene load_steps=2 format=2]
+
+[ext_resource path="res://benchmarks/BunnymarkV1DrawTexture/d/BunnymarkV1DrawTexture.gdns" type="Script" id=1]
+
+[node name="Test" type="Node2D"]
+
+script = ExtResource( 1 )
+_sections_unfolded = [ "Material", "Visibility", "Z" ]
+
+
```
chippmann: Is this one needed?

samples/benchmark/bunnymark/export_presets.cfg
```diff
@@ -0,0 +1,24 @@
+[preset.0]
+
+name="Linux/X11"
+platform="Linux/X11"
+runnable=true
+custom_features="release"
+export_filter="all_resources"
+include_filter=""
+exclude_filter=""
+export_path="/home/carter/code/godot3-bunnymark/export/Bunnymark.x86_64"
```
chippmann: export presets should not be committed as they contain absolute file paths

samples/benchmark/bunnymark/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,135 @@
...
+
+buildscript {
+    repositories {
+        mavenLocal()
+        maven("https://dl.bintray.com/utopia-rise/kotlin-godot")
+        jcenter()
+    }
+    dependencies {
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.61")
+        classpath("org.godotengine.kotlin:godot-gradle-plugin:1.0.1")
```
chippmann: I would use the beta versions `0.1.0-3.2`

samples/benchmark/bunnymark/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+fun configureTargetAction(kotlinTarget: @ParameterName(name = "target") org.jetbrains.kotlin.gradle.plugin.KotlinTarget) {
+    kotlinTarget.compilations.getByName("main") {
+        if (this is org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeCompilation) {
+            println("Configuring target ${target.name}")
+            target.compilations.all {
+                dependencies {
+                    implementation("org.godotengine.kotlin:godot-library:1.0.0")
+                    implementation("org.godotengine.kotlin:annotations:0.0.2")
```
chippmann: Versions

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV1DrawTexture.kt
```diff
@@ -0,0 +1,97 @@
...
+
+@RegisterClass("benchmarks/BunnymarkV1DrawTexture/kot")
+class BunnymarkV1DrawTexture: Node2D() {
+
+    data class Bunny(var position:Vector2, var speed: Vector2)
+
+    private var bunnies = ArrayList<Bunny>()
+    private var grav = 500
+    private var bunnyTexture = Texture from ResourceLoader.load("res://images/godot_bunny.png")
+    private val rng = RandomNumberGenerator()
```
chippmann: Variable name is not self explanatory. I would prefer `randomNumberGenerator`.
But you can leave it if you prefer this.

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV1DrawTexture.kt
```diff
@@ -0,0 +1,97 @@
+package godot.samples.benchmark.bunnymark
```
chippmann, line 1: Format the script once with intellij's auto formatter :-)

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV1Sprites.kt
```diff
@@ -0,0 +1,91 @@
...
+
+@RegisterClass("benchmarks/BunnymarkV1Sprites/kot")
+class BunnymarkV1Sprites: Node2D() {
+
+    private data class Bunny(var sprite: Sprite, var speed: Vector2)
+
+    private var bunnies = ArrayList<Bunny>()
+    private var grav = 500
+    private var bunnyTexture = Texture from ResourceLoader.load("res://images/godot_bunny.png")
+    private val rng = RandomNumberGenerator()
```
chippmann: Again variable name. But as said: optional

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV1Sprites.kt
```diff
@@ -0,0 +1,91 @@
+package godot.samples.benchmark.bunnymark
```
chippmann, line 1: auto format the script once :-)

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV2.kt
```diff
@@ -0,0 +1,98 @@
...
+import godot.*
+import org.godotengine.kotlin.annotation.RegisterClass
+import org.godotengine.kotlin.annotation.RegisterFunction
+
+@RegisterClass("benchmarks/BunnymarkV2/kot")
+class BunnymarkV2: Node2D() {
+
+
+    private var grav = 500
+    private var bunny_speeds = ArrayList<Vector2>()
```
chippmann: variable names should be camelCase when not exported

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV2.kt
```diff
@@ -0,0 +1,98 @@
...
+    private var label = Label()
+    private var bunnies = Node2D()
+    private var bunnyTexture = Texture from ResourceLoader.load("res://images/godot_bunny.png")
+    private val rng = RandomNumberGenerator()
```
chippmann: Again variable name. But as said: optional

samples/benchmark/bunnymark/kotlin/src/main/kotlin/godot/samples/benchmark/bunnymark/BunnymarkV2.kt
```diff
@@ -0,0 +1,98 @@
+package godot.samples.benchmark.bunnymark
```
chippmann, line 1: auto format the script once

samples/benchmark/bunnymark/run_benchmarks.sh
```diff
@@ -0,0 +1,14 @@
...
+        for language_path in $benchmark_path/*; do
+			if [ -d $language_path ]
+			then
+				language=$(basename $language_path)
+				echo $benchmark $language
+				export/Bunnymark.x86_64 --bench=$benchmark --lang=$language
+			fi
+        done
+    fi
+done
```
chippmann: I don't think this is needed.
If you think otherwise, you should add a script for windows as well

chippmann (changes requested):
Thanks a lot for the implementation of the benchmark!

chippmann:
Maybe we could think of putting the benchmark results in a spreadsheet and update it on each release, so we see performance changes over time?
Maybe we could even automate this somehow.

samples/benchmark/bunnymark/Test.tscn
piiertho: I guess it is, as it seems to be the benchmark scene ?

samples/benchmark/bunnymark/kotlin/README.md
```diff
@@ -0,0 +1,3 @@
+# Games sample
+
+Use `gradlew build` to build this sample
```
piiertho: Maybe add a description of how to build for a specific platform ?

samples/benchmark/bunnymark/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,135 @@
...
+plugins {
+    id("org.jetbrains.kotlin.multiplatform") version ("1.3.61")
+}
+
+apply(plugin = "godot-gradle-plugin")
+
+repositories {
+    mavenLocal()
+    maven("https://dl.bintray.com/utopia-rise/kotlin-godot")
+    jcenter()
```
piiertho: It's missing exclusion for kotlinx coroutines module. Even if we don't use them here I think we should still have same template for our samples so that user don't get lost.

piiertho (changes requested):
I agree with @chippmann request. Also I commented for few changes.

samples/benchmark/bunnymark/Test.tscn
chippmann: true i didn't pay enough attention

raniejade:
Was able to port he DrawTexture benchmark to my binding and it is looking grim. I got a lower number (1000). Used linux's perf tool and have identified the following as "expensive":

- calling c methods (specially the ones returning `CValues`, there's a hidden step here - moving the value from the stack into kotlin's memory)
- `MemScope.ptr`/`getPointer` (similar to above but the opposite direction, kotlin memory -> native heap)
- KString -> c string conversions (mainly on method_bind calls, caching them would solve the problem)
- garbage collection (I'll have to look at this more)

CedNaru:
There is still a lot of things we can optimize. 
But if what cost the most perf is the memory context switch between C and Kotlin. There is not much we can do until K/N perfs improve.
We also have a lot of memory leaks on our side because we can't call the destructor of the native object when the koltin wrapper is freed by the GC.

chippmann:
@raniejade @CedNaru @piiertho before we now do performance optimizations on each of our bindings, maybe now would be a good time to speak together about the future of our projects.
I think it's too early to speak of merging but surely we can do some sort of know how transfer and maybe work a bit closer together to see in which direction each binding is heading, what works, what does not and maybe even discuss some topics directly together across our projects?

piiertho:
@chippmann I totally agree with you !
I wanted to send a mail to @raniejade today, but as you are already here, I think we can start discuss here.
I think we should have a discord call to organize ourself :)
I'd really like we share our experiences, what will make a future merge easier to do.

raniejade:
I'm quite happy to do some knowledge sharing. What timezones are you guys at?

chippmann:
GMT +1 (CET) at least we maintainers.

raniejade:
I'm in the land down under :joy: UTC + 11 (AEDT). It will probably be hard to set a meeting :). Let's just discuss here.

Wavesonics:
UTC -8 (PST) here

chippmann (approved):
I'm fine with it now.
Thanks a lot for the implementation!

piiertho (approved):
Thanks !

samples/benchmark/bunnymark/kotlin/build.gradle.kts
```diff
@@ -0,0 +1,150 @@
...
+                }
+            }
+            if (project.hasProperty("iosSigningIdentity") && this.target.name == "iosArm64") {
+                tasks.build {
+                    doLast {
+                        exec {
+                            commandLine = listOf("codesign", "-f", "-s", iosSigningIdentity, "build/bin/iosArm64/releaseShared/libkotlin.dylib")
+                        }
+                        exec {
+                            commandLine = listOf("install_name_tool", "-id", "@executable_path/dylibs/ios/libkotlin.dylib", "build/bin/iosArm64/releaseShared/libkotlin.dylib")
```
piiertho, line 130: Same as the other PR (#59 ). I did not see it in first place. I think I did not see it in buildType PR.
@chippmann do you think we can go with hard coded debug or release for ios signing here and make another PR to add this to all projects samples ?

piiertho (changes requested):

samples/benchmark/bunnymark/kotlin/build.gradle.kts
chippmann: Yes i think so. As we overlooked it previously, we can merge it for now and fix it in all samples in a separate PR.

piiertho (approved):
Ok for me !

PR #61, opened 2020-03-07
Fix bug in entry generation where signal parameter strings where not escaped

chippmann:
This fixes a bug where signal parameter name strings were not escaped.
ex: `registerSignal("godot.samples.coroutines.Ball", "move", arrayOf(step to Variant.Type.VECTOR2), arrayOf(Variant(godot.core.Vector2())))`
`step` was missing the `"`.
Now it's correctly escaped and becomes `registerSignal("godot.samples.coroutines.Ball", "move", arrayOf("step" to Variant.Type.VECTOR2), arrayOf(Variant(godot.core.Vector2())))`


piiertho (approved):
Thanks for bug fix !
Will go in beta !

Issue #52, opened 2020-03-01
Implement the global functions of GDScript in the GD singleton


CedNaru:
**Describe the problem or limitation you are having in your project:**

When converting script from GDScript to Kotlin, there were a lot of functions I couldn't use out of the box.
Some are straightforward, some need to be locally rewritten, some need to be imported from external Kotlin package.

**Describe how this feature / enhancement will help you overcome this problem or limitation:**

The idea is to implement the different function of GdScript described here.
https://docs.godotengine.org/en/3.2/classes/class_@gdscript.html

Use cases:
- Math Functions
- Files Loading
- Random numbers generation
- Shortcuts to many Godot objects without the need to instantiate them.

**Describe implementation detail for your proposal (in code), if possible:**
These functions could be implementented in the GD singleton just like the Print function is today.

chippmann:
I personally don't like the GD singleton and want to get rid of it wherever we can.
Personally I'm for global functions to implement your proposal.

chippmann:
I'll implememt this together with #39

chippmann:
@CedNaru I assigned you this (a little bit older) issue as you are working on it currently in the master merge.
Feel free to close it otherwise if you don't see it fit.

CedNaru:
Okay:
A reminder that we decided to keep the gd singleton in order to have a safe namespace as the godot.core package is getting a bit too much bloated now.

Issue #39, opened 2020-02-22
Missing duplicate (with deep_copy bool) function for Dictionary


2shady4u:
Hello!

**Duplicate(deep = false)**
I'm currently missing the duplicate function (with deep_copy functionalities) in the godot.core.Dictionary.

**keys and values?**
Also I'm not really sure about the implementation of keys and values as they currently seem
much too close to being actual fields in the Dictionary?
Like for example if I have following Dictionary dict:
```
{
"keys" : ["pear", "banana", "apple"],
"values": ["diamond", "pearl", "amethyst"]
}
```
I would expect dict.keys to give me a GDArray of ["pear", "banana", "apple"] instead of a GDArray of ["keys", "values"]? Currently this isn't the case and this might confuse adapters coming from GDScript. (I am aware that you can still use dict["keys"] and that will work... but the confusion can be avoided!)

**list\<String> and array\<String> to PoolStringArray**
There's a lot of other functionalities missing but having a conversion between list\<String> and PoolStringArray (as well as other lists<>/array<> to their respective Godot arrays) would be great to have.

chippmann:
I implemented the missing duplicate function in the branch: https://github.com/utopia-rise/godot-kotlin/tree/feature/add_missing_functions_to_core_classes

Regarding Dictionary keys:
It already works like in GDScript as far as i can see.
In your example the keys of the Dictionary are in fact `keys` and `values`. And the values are `["pear", "banana", "apple"]` and `["diamond", "pearl", "amethyst"]`.
That works the same in GDScript?
What you basically ask for is a Map which does not exist in GDScript.
Or do i not understand you correctly?

Good idea about the list conversion. I look that i can implement that. Are there some other missing functionalities you want to see or missing functions you came across?

2shady4u:
I wanted to try this myself :'( I don't get why it doesn't work.

chippmann:
> I wanted to try this myself :'( I don't get why it doesn't work.

Oh sorry ^^ Should i assign it to you?

2shady4u:
It's ok, I would just like to know the issue with my forked branch though. I would have made a PR days ago, if it just worked instead of crashed.

Will make an example to clarify the issue with keys and values.

2shady4u:
It's ok, I would just like to know the issue with my forked branch though. I would have made a PR days ago, if it just worked instead of crashed.

Will make an example to clarify the issue with keys and values.

piiertho:
[...]

PR #63, opened 2020-03-07
Feature/registration default value improvement

### LINKED ISSUE #53: Improve default value setting in RegisterProperty and RegisterSignal annotations
Currently it's very cumbersome and not typesafe to define default values. On top of that, the user in most cases defines a default value in code anyways for properties. And for signals it wpould be easy as well.

The user can use kotlin default value features for registering signals and normal initialisation for properties instead of an usafe string.

The user would write just this to register a property with a default value:
```
@RegisterProperty(visibleInEditor = false)
val someProperty = "I'm an awesome String!"
```
and this for Signals with default values:
```
@RegisterSignal
fun someSignal(firstArgument: Boolean = false, secondArgument: String = "coolStuff") {}
```

We need to find a way to get the values of the properties and parameters inside the entry generation code generation in a typesafe way.

Yes we could just use the current string version but it's ugly and not typesafe at all

Entry generation is part of the core project.

chippmann:
This resolves #53.
This improves the default value registration for properties by leveraging the psi system to read the default value from code.
A user can now write:
```
@RegisterProperty(true)
var playerBodyNodePath = NodePath("Entities/Player/PlayerBody")
```
To register properties. The default value will in this example be `godot.core.NodePath("Entities/Player/PlayerBody")`.

Also this removes the possibility of defining default values for signals.
This is done because it didn't work in the first place and it seems not supported anymore anyways. I found no way to make it work and the cpp bindings for example don't seem to support it either. The same can be achieved anyways through binds in the `connect` method.
I someone has an idea of how tho make this work, I'm open for suggestions.
What i tried:
- Passing the individual default arguments one by one to the vararg `default_args` in the `register_signal` function.
- Passing all of them together to the `default_arg` parameter as a GDArray

Both ways if the signal was emitted with no arguments, the `args` value we get from godot is null.

I opted for removing this feature as i also didn't find it in the cpp binding, nor did i find any sort of documentation about it.

### DIFF
--- a/API_DIFFERENCES.md
+++ b/API_DIFFERENCES.md
@@ -54,7 +54,7 @@ To use `yield` keyword in your program **library extension** is needed.
                 }
                 target.compilations.all {
                     dependencies {
-                        implementation("org.godotengine.kotlin:godot-library-extension:0.1.0-3.2") // <-- This one here
+                        implementation("org.godotengine.kotlin:godot-library-extension:0.1.1-3.2") // <-- This one here
                     }
                 }
             } else {
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -59,7 +59,7 @@ buildscript {
     }
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.61")
-        classpath("org.godotengine.kotlin:godot-gradle-plugin:0.1.0-3.2")
+        classpath("org.godotengine.kotlin:godot-gradle-plugin:0.1.1-3.2")
     }
 }
 
@@ -198,8 +198,8 @@ fun configureTargetAction(kotlinTarget: @ParameterName(name = "target") KotlinTa
             println("Configuring target ${target.name}")
             target.compilations.all {
                 dependencies {
-                    implementation("org.godotengine.kotlin:godot-library:0.1.0-3.2")  // or implementation("org.godotengine.kotlin:godot-library-extension:0.1.0-3.2") if you want coroutines like yield
-                    implementation("org.godotengine.kotlin:annotations:0.1.0-3.2")
+                    implementation("org.godotengine.kotlin:godot-library:0.1.1-3.2")  // or implementation("org.godotengine.kotlin:godot-library-extension:0.1.1-3.2") if you want coroutines like yield
+                    implementation("org.godotengine.kotlin:annotations:0.1.1-3.2")
                 }
             }
             if (project.hasProperty("iosSigningIdentity") && this.target.name == "iosArm64") {
@@ -249,7 +249,7 @@ buildscript {
     }
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.61")
-        classpath("org.godotengine.kotlin:godot-gradle-plugin:0.1.0-3.2")
+        classpath("org.godotengine.kotlin:godot-gradle-plugin:0.1.1-3.2")
     }
 }
 
@@ -353,8 +353,8 @@ fun configureTargetAction(kotlinTarget: @ParameterName(name = "target") org.jetb
             println("Configuring target ${target.name}")
             target.compilations.all {
                 dependencies {
-                    implementation("org.godotengine.kotlin:godot-library:0.1.0-3.2") // or implementation("org.godotengine.kotlin:godot-library-extension:1.0.0") if you want coroutines like yield
-                    implementation("org.godotengine.kotlin:annotations:0.1.0-3.2")
+                    implementation("org.godotengine.kotlin:godot-library:0.1.1-3.2") // or implementation("org.godotengine.kotlin:godot-library-extension:1.0.0") if you want coroutines like yield
+                    implementation("org.godotengine.kotlin:annotations:0.1.1-3.2")
                 }
             }
             if (project.hasProperty("iosSigningIdentity") && this.target.name == "iosArm64") {
--- a/REGISTRATION.md
+++ b/REGISTRATION.md
@@ -108,7 +108,7 @@ val nodePathAsString = "Some/Path/To/A/Node"
 @RegisterProperty(true)
 var aProperty: Boolean = NodePath(nodePathAsString)
 ```
-If you try to do that, you will get a unknown reference exception!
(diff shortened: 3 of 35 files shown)


raniejade:
Interesting approach, but what happens when you have:

```kotlin
val pathStr = "Entities/Player/PlayerBody"

@RegisterProperty
val path = NodePath(pathStr)
```

chippmann:
@raniejade At the moment this does not work. It would leave to an unresolved reference.
I should probably mention this in the docs as well.
It is possible to follow the reference but i didn't do this for now.

raniejade:
I don't think the reference should be followed because you might be accessing a property of `this` which is not available during registration time.

```
@RegisterProperty
val path = NodePath(getSomePath())
```

chippmann:
True.
But i should check in the future if the element at that position is a valid assignment. But for now i think its ok if i just mention this in the docs.

piiertho (approved):
Thanks for the enhancement and debug ! :)

piiertho:
> True.
> But i should check in the future if the element at that position is a valid assignment. But for now i think its ok if i just mention this in the docs.

oops, looked to changes before looking to discussion. Can you add this to documentation for the moment, then it should be ok for now. Thanks @raniejade for highlighting this problem.

piiertho (changes requested):
Add missing doc.
Also, now that #61 is merged, maybe we can have longer reflexion on problem on accessing variables from this at registration time ?

chippmann:
@piiertho sorry i don't get what you mean

chippmann:
Damn i also managed to push `platform=linux` in the samples. I'll remove that as well.

piiertho:
@chippmann i was meaning that maybe we can take more time to think about the problem @raniejade mentionned.

chippmann:
@piiertho I see. I'm generally fine with this idea. But i would like to merge it first anyways as it removes the not working signal default arguments we for now have included.
But i could extract that in a separate PR so we don't need to merge this one for now if you want. That would be fine as well i guess.
Anyways i will stale my work on #39 and #52 until i completely improved this.

chippmann:
@piiertho and @raniejade 
My current implementation correctly detects the following wrongly defined registered properties:
```
val blubb = ""
@RegisterProperty(true)
var test1 = NodePath(blubb)

@RegisterProperty(true)
var test2 = Vector3::x

private val ref = 300
@RegisterProperty(true)
var test3 = this.ref

private fun getRef() = 300
@RegisterProperty(true)
var function = this.getRef()

lateinit var velocity: Vector2
@RegisterProperty(true)
var test4 = velocity.x
```

Do you have any other mistakes in inititalisation the user might be able to do?
lateinits, val, and by lazy was already handled

piiertho:
@chippmann what about others delegate properties ? Does that works ?

piiertho:
Mmmh maybe that's a stupid question, as we don't know how custom delegates works...

chippmann:
It should work. But to test it, you could make an example how you would write it and i try it as well. Regarding delegates i just tried by lazy.

raniejade:
We need to make sure that the exposed property is `Variant` compatible (i.e can be converted to a `Variant`).

I initially disliked the approach as it can cause unexpected errors.
- referencing values from `this` which is not available during registration time
- not all types can be exposed to Godot.

But I think with some IDE inspections we can easily educate users about this caveat.

On my binding I go around the caveats above by:

> - referencing values from `this` which is not available during registration time

registration happens manually and outside of `this` context. So the default value can't reference any values from `this`.

> - not all types can be exposed to Godot.

By using delegates and only implemented it for types that are `Variant` compatible. This means that I don't have to rely on a custom compiler plugin to do the checks for me - the kotlin compiler does it out of the box.

I wonder if we can come up with a compromise of the two approach?

```kotlin
@RegisterProperty
val myProperty by property<Int> { ... }
```

Pros:
- automatic registration
- no need for custom compiler plugin to check if type can be exposed

Cons:
- The issue with referencing `this` is still there.

With the mentioned cons above, we can just do the check in the compiler plugin. i.e check that the lambda is `non-capturing`. The kotlin does the same thing for `Worker.execute` and `staticCFunction`. We probably can even file a language proposal to support our use-case.

Something like:

```kotlin
fun foo(m: static () -> Unit)
```

Another advantage of this approach is that we can provide type hints in a type-safe and compatible way. When I say `compatible` it means a hint can only be used for a specific type (https://docs.godotengine.org/en/3.2/getting_started/scripting/gdscript/gdscript_exports.html). The `String` type for example can be used as filesystem paths:

excerpt:
```gdscript
# String is a path to a file.
export(String, FILE) var f
# String is a path to a directory.
export(String, DIR) var f
# String is a path to a file, custom filter provided as hint.
export(String, FILE, "*.txt") var f

# Using paths in the global filesystem is also possible,
# but only in scripts in "tool" mode.

# String is a path to a PNG file in the global filesystem.
export(String, FILE, GLOBAL, "*.png") var tool_image
# String is a path to a directory in the global filesystem.
export(String, DIR, GLOBAL) var tool_dir

# The MULTILINE setting tells the editor to show a large input
# field for editing over multiple lines.
export(String, MULTILINE) var text
```

We can do:

```kotlin
val globalPath by property<String>(hint = PropertyHint.file(global = true)) { ... }
val localPathWithFilter by property<String>(hint = PropertyHint.file(filter = "*.png")) { ... }

// sensible defaults, hint = PropertyHint.string()
val onelineString by property<String> { ... }
// multiline string
val multilineString by property<String>(hint = PropertyHint.string(true)) { ... }
```

chippmann:
Interesting idea. I quite like it, but i need to think a bit more about it. One immediate question arises though: how do you define the default value when using:
```kotlin
@RegisterProperty
val myProperty by property<Int> { ... }
```

chippmann:
Of course we could define it as an additional param but i personally don't quite like it.

raniejade:
It's the return value of the lambda.

```kotlin
val myProperty by property<Int> { 1 }
```

It might be easier to check via the compiler plugin if the lambda is capturing or not. (maybe via `LambdaUtil.isCapturingLambda(...)`). What we want here is a non-capturing one.

chippmann:
Closed in favor of new implementation in master-merge for new binding

PR #76, opened 2020-03-29
Implement gradle plugin

raniejade:
This is it how it currently looks:

```kotlin
plugins {
  id("com.utopia-rise.godot-kotlin")
  kotlin("multiplatform") version "1.3.71"
}

godot {
  debug.set(true)
  cleanupGeneratedFiles.set(true)

  configureTargets {
    compilations.getByName("main") {
      cinterops {
        // create cinterop here
      }
    }
  }
}
```

The plugin will automatically configure the `multiplatform` plugin and setup the appropriate targets based on `godot.platforms` property. A source set `godotMain` is created and linked to the created target's source set `targetSourceSet.dependsOn(godotMain)`. This means that `kotlin.mpp.enableGranularSourceSetsMetadata` must be set to `true` in `gradle.properties` to get proper completion support.

A dummy target is also created based on the current operating system, which is basically used to generate the entry files.

The plugin will also create a build task per target. (i.e `buildLinuxX64`, `buildWindowsX64`, etc ...) which can be used to build the shared library.


samples/mini-games/settings.gradle.kts
```diff
@@ -0,0 +1,13 @@
+includeBuild("../../") {
```
raniejade, line 2: @chippmann I know you don't like this, but this will help us iterate faster. Let's decide in removing it once we have a stable version.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotExtension.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.gradle.kotlin.dsl.property
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+
+open class GodotExtension(objects: ObjectFactory) {
+    val debug = objects.property<Boolean>()
+    val cleanupGeneratedFiles = objects.property<Boolean>()
+    val gdnsDir = objects.fileProperty()
+    val platforms = objects.listProperty<Platform>()
+
+    internal val entrySourceDir = objects.directoryProperty()
```
raniejade, line 14: Set to `build/godot-entry/`, this can be used by the kotlin subplugin.

samples/mini-games/settings.gradle.kts
chippmann: Agree. Let's discuss this again once the merge is done :-) For now though i'm fine with it.

build.gradle.kts
```diff
@@ -1,4 +1,6 @@
 subprojects {
+    group = "com.utopia-rise"
+    version = "0.1.0"
```
chippmann, line 3: I would move the version declaration here:
image
To have all at the same place and ensure it is not missed when bumping a version.

plugins/godot-gradle-plugin/build.gradle.kts
```diff
@@ -1,5 +1,30 @@
...
+            displayName = "Gradle plugin for godot-kotlin"
+            implementationClass = "godot.gradle.GodotPlugin"
+        }
+    }
+    isAutomatedPublishing = false
+}
+
+pluginBundle {
+    website = "https://github.com/raniejade/godot-kotlin"
+    vcsUrl = "https://github.com/raniejade/godot-kotlin.git"
```
chippmann: Still links to the wrong webiste ;-)

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -0,0 +1,111 @@
...
+            val godot = project.extensions.create<GodotExtension>("godot", project.objects)
+            setupExtensionDefaults(project, godot)
+            setupKotlinPlugin(project, mpp, godot)
+        }
+    }
+
+    private fun setupExtensionDefaults(project: Project, godot: GodotExtension) {
+        with(godot) {
+            platforms(*Platform.values())
+            debug.set(false)
```
chippmann: Don't we want to set the debug target as default? I guess when one is starting using our binding, he starts developing and most of the time uses the debug build then?

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -0,0 +1,111 @@
...
+            val buildTask = project.tasks.getByName("build")
+
+            // create the targets and connect it to the main source set
+            godot.platforms.get().forEach { platform ->
+                val target = when (platform) {
+                    Platform.LINUX_X64 -> mpp.linuxX64("linuxX64")
+                    Platform.WINDOWS_X64 -> mpp.mingwX64("windowsX64")
+                    Platform.OSX_X64 -> mpp.macosX64("macosX64")
+                    Platform.ANDROID_X32 -> mpp.androidNativeArm32("androidX32")
+                    Platform.ANDROID_X64 -> mpp.androidNativeArm64("androidX64")
```
chippmann: These are the arm targets not the X targets right? So naming is wrong? Also then the "real" X targets are missing?

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -0,0 +1,111 @@
...
+                    Platform.IOS_X64 -> mpp.iosX64("iosX64")
```
chippmann, line 74: For iOS the arm target is missing?

raniejade:
I've included some files in `.idea` directory.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/Platform.kt
```diff
@@ -0,0 +1,10 @@
+package godot.gradle
+
+enum class Platform {
```
chippmann, line 3: As said in `GodotPlugin.kt` some targets seem to be missing

samples/mini-games/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-bin.zip
```
chippmann: also for the samples i woild set the flavour to `all`

samples/mini-games/settings.gradle.kts
```diff
@@ -0,0 +1,13 @@
...
+    dependencySubstitution {
+        substitute(module("com.utopia-rise:godot-gradle-plugin")).with(project(":godot-gradle-plugin"))
+    }
+}
+
+pluginManagement {
+    resolutionStrategy.eachPlugin {
+        when (requested.id.id) {
+            "com.utopia-rise.godot-kotlin" -> useModule("com.utopia-rise:godot-gradle-plugin:0.1.0")
```
chippmann, line 12: Also here should the version not go into the versions file in the buildSrc?

chippmann (approved):
Really awesome work! Some minor details need to be resolved though.

chippmann:
Shit wrong button xD I meant needs work xD

.idea/misc.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="JavaScriptSettings">
+    <option name="languageLevel" value="ES6" />
+  </component>
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" project-jdk-name="11" project-jdk-type="JavaSDK" />
```
chippmann: Shouldn't this be set to java 8 for compability?

chippmann (changes requested):

build.gradle.kts
raniejade: This is the project version, that thing is reserved for third party dependencies.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
raniejade: Yep, makes sense.

samples/mini-games/settings.gradle.kts
raniejade: It's not accessible here, the `0.1.0` is here so that gradle won't complain. Basically it is just a placeholder value to make gradle happy.

.idea/misc.xml
raniejade: :ok_hand:

build.gradle.kts
chippmann: Ah then i guess we misundersood each other. I was thinking of defining all versions in that file. Dependencies and own versions. The own versions i thought it makes it easy to bump versions and don't forget about one in one of the modules.
But your right about hird party dependencies. Maybe it makes sense to make two such file then? One for our versions and one for third parties?

samples/mini-games/settings.gradle.kts
chippmann: I see. Ok then ^^

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/Platform.kt
raniejade: Yeah, nice catch. I forgot about arm and non-arm variants for mobile.

build.gradle.kts
raniejade: We will probably just have a single version, why not just keep it in the root `build.gradle.kts` file?

plugins/godot-gradle-plugin/build.gradle.kts
raniejade: fixed :) copy-pasta errors, sorry! XD

raniejade (commented):
@chippmann should be ready for the round of reviews :)

build.gradle.kts
chippmann: fair enough :-)
I'm fine with it

plugins/godot-gradle-plugin/build.gradle.kts
chippmann: mmmh pasta xD I'm getting hungry xD

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -0,0 +1,114 @@
...
+            val buildTask = project.tasks.getByName("build")
+
+            // create the targets and connect it to the main source set
+            godot.platforms.get().forEach { platform ->
+                val target = when (platform) {
+                    Platform.LINUX_X64 -> mpp.linuxX64("linuxX64")
+                    Platform.WINDOWS_X64 -> mpp.mingwX64("windowsX64")
+                    Platform.OSX_X64 -> mpp.macosX64("macosX64")
+                    Platform.ANDROID_X32 -> mpp.androidNativeX64("androidX32")
+                    Platform.ANDROID_X64 -> mpp.androidNativeX64("androidX64")
```
chippmann: Both are X64 now ;-)

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/Platform.kt
```diff
@@ -0,0 +1,13 @@
...
+    LINUX_X64,
+    WINDOWS_X64,
+    OSX_X64,
+    ANDROID_X32,
```
chippmann: We will need to check if we can even compile this. We hadn't had any 32bit targets because on 32bit we ran into out of memory exceptions because we were using to much ram for 32bit hosts to compile the project. But if i remember correctly one can build for 32bit android on x64 anyways but i'm not sure.
Anyways i'm fine with it for now.

chippmann (changes requested):
One small change need to be done.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
raniejade: :man_facepalming:

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/Platform.kt
raniejade: We could just drop the `32` bit targets for now and just add them later on when someone asks for it. Wdyt?

samples/mini-games/settings.gradle.kts
```diff
@@ -0,0 +1,13 @@
+includeBuild("../../") {
+    dependencySubstitution {
+        substitute(module("com.utopia-rise:godot-gradle-plugin")).with(project(":godot-gradle-plugin"))
+    }
+}
+
+pluginManagement {
```
piiertho, line 9: Maybe we can add comments on placeholders line ?

samples/mini-games/settings.gradle.kts
piiertho, line 2: same, maybe we can add comment here :)

piiertho (changes requested):

chippmann (approved):
Thanks for the changes.

piiertho (approved):
Thanks a lot !

PR #78, opened 2020-03-29
Implement registration annotations and temporary `RPCMode` enum

### LINKED ISSUE #68: Implement registration annotations
Implement all needed registration annotations. This includes:  
- @RegisterClass
- @RegisterProperty
- @RegisterFunction
- @RegisterSignal

These annotations should have the following capabilities (arguments):
- **@RegisterClass**  
None
- **@RegisterProperty**  
visibleInEditor: Boolean (Optional)  
rpcMode: RPCMode (Optional)
- **@RegisterFunction**  
rpcMode: RPCMode (Optional)
- **@RegisterSignal**  
None

**Hints:**
The RPCMode's and the Property Hint's will be implemented as separate annotations rather than as arguments as they are now.  
If the necessary RPCMode enum does not exist yet, add an empty enum class and add a TODO in it.

chippmann:
The RPCMode enum needs to be properly implemented later on.
It was just implemented so it can be added as an argument to
the annotations.

Also the build gradle for the `godot-library` needs proper implementation once the development starts there.

This resolves #68

### DIFF
--- a/godot-kotlin/godot-library/build.gradle.kts
+++ b/godot-kotlin/godot-library/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+
 plugins {
     kotlin("multiplatform")
 }
@@ -15,7 +17,7 @@ kotlin {
 
     val nativeMain = sourceSets.create("nativeMain")
 
-    targets.withType<org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget> {
+    targets.withType<KotlinNativeTarget> {
         compilations.getByName("main").defaultSourceSet { dependsOn(nativeMain) }
     }
 }
--- a/godot-kotlin/godot-library/build.gradle.kts
+++ b/godot-kotlin/godot-library/build.gradle.kts
@@ -1,3 +1,21 @@
 plugins {
     kotlin("multiplatform")
 }
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+//TODO: this needs to be properly configured! This is just a basic setup to be able to implement the annotations
+kotlin {
+    linuxX64()
+    mingwX64()
+    macosX64()
+
+    val nativeMain = sourceSets.create("nativeMain")
+
+    targets.withType<org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget> {
+        compilations.getByName("main").defaultSourceSet { dependsOn(nativeMain) }
+    }
+}
--- a/godot-kotlin/godot-library/build.gradle.kts
+++ b/godot-kotlin/godot-library/build.gradle.kts
@@ -4,11 +4,6 @@ plugins {
     kotlin("multiplatform")
 }
 
-java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
-}
-
 //TODO: this needs to be properly configured! This is just a basic setup to be able to implement the annotations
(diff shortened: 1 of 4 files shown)


godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -1,3 +1,21 @@
...
+
+//TODO: this needs to be properly configured! This is just a basic setup to be able to implement the annotations
+kotlin {
+    linuxX64()
+    mingwX64()
+    macosX64()
+
+    val nativeMain = sourceSets.create("nativeMain")
+
+    targets.withType<org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget> {
```
raniejade: add a qualified import on top?

godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -1,3 +1,21 @@
 plugins {
     kotlin("multiplatform")
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
```
raniejade: Do we need this here? We are not targeting the JVM so I guess this is not required?

raniejade (changes requested):
A couple of comments, but otherwise LGTM!

godot-kotlin/godot-library/build.gradle.kts
chippmann: We had once the problem that a user build the lib locally with java 13 because nothing was set here, but in his game he specified java 8 in the IDE which lead the game to not be able to compile.
This solved it in that case.
Do you have a better idea to enforce java 8 compatibility when built locally?

godot-kotlin/godot-library/build.gradle.kts
chippmann: Sure I'll change it

godot-kotlin/godot-library/build.gradle.kts
raniejade: That's super weird. This setting should only affect a jvm compilation which in this case we don't have any. Is there a ticket somewhere I could look at? Or was it in discord?

chippmann: It was on discord. But maybe you are right. Because we had in the old binding a jvm target as the annotations were a multiplatform module which had a jvm target for the entry generator.
But we will not have that so i guess i can remove it :thinking:

raniejade: Let's add it later on once we encounter it again :)

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/registration/RPCMode.kt
```diff
@@ -0,0 +1,14 @@
+package godot.registration
+
+//TODO: temporary enum class! Needs to be properly implemented! Implementation was done to be able to implement register annotations
+enum class RPCMode {
+    Disabled,
```
raniejade: We should be consistent on how we capitalize enum values. On my PR I had `Platform` enum values all caps. Should we do the same here? Or maybe the reverse?

raniejade: The stdlib way is having it all caps and underscore separated when needed.

raniejade (commented):
One last comment :)

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/registration/RPCMode.kt
chippmann: yeah here as well: https://kotlinlang.org/docs/reference/enum-classes.html
I'll change it to all caps :-)

raniejade (approved):
LGTM!

PR #79, opened 2020-03-29
Implement rpc annotations

### LINKED ISSUE #70: Implement RPCMode annotations
All enums defined in RPCMode should also exist as annotations:
- Remote
- Disabled
- Sync
- Master
- Puppet
- RemoteSync
- MasterSync
- PuppetSync

The annotation do not have parameters.

chippmann:
This should be merged after #78 .

These are the RPCMode annotations a user can use instead of the RPCMode enum argument inside the `@RegisterProperty` and the `@RegisterFunction` annotations.

Resolves #70 .

### DIFF
--- a/godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/RpcAnnotation.kt
+++ b/godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/RpcAnnotation.kt
@@ -1,0 +1,33 @@
+package godot.annotation
+
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Remote
+
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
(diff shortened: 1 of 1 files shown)


raniejade:
Why do we need two ways to specify the rpc mode? I would prefer to just have one, less thing we have to support.

chippmann:
True. The idea was that the user can do what he prefers. Consider this case:

```
@IntRange(0, 100, 2)
@RegisterProperty(true, RPCMode.Master)
var foo: Int = 2
```
vs:
```
@Master
@IntRange(0, 100, 2)
@RegisterProperty(true)
var foo: Int = 2
```

If even more annotations wpould be added in the future we would have even more lines

raniejade:
I think we should only support one, I personally prefer the enums as it clearly shows the relationship of `RpcMode` and `RegisterMethod`/`RegisterFunction`. In my experience with maintaining libraries having a single way to do things is beneficial in the long run for maintainers. It also simplifies the entry generator as it only needs to check a single thing. It also simplifies documentation :)

chippmann:
Ok I'm fine with this. I'll then close this PR and the corresponding issue. We can still add them if we decide to do this way later on.

PR #80, opened 2020-03-29
Implement property hint annotations

### LINKED ISSUE #69: Implement propertyType hint annotations
Implement all needed propertyHint annotations. Like for example:  
- @IntRange

These annotations should have their arguments optional and in the correct type. For example:
- @IntRange
start: Int  
end: Int  
step: Int

**Hints:**
A hint on what should and could be implemented: http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_exports.html#doc-gdscript-exports

chippmann:
I hope i have catched all property type hint's with the correct arguments.

I looked at the godot source code [here](https://github.com/godotengine/godot/blob/5f1107aa30295e686be6f41cb6d17fc2cff1e036/editor/editor_properties.cpp) to see which type hints exist, on what types they apply and what arguments they might take. Also a helpful link to the godot documentation about exporting variables [here](http://docs.godotengine.org/en/latest/getting_started/scripting/gdscript/gdscript_exports.html#doc-gdscript-exports).

A note. As far as I know, one cannot define for an annotation on what property types a user can apply those. So i guess for now we will have to make those checks in the entry-generator at compile time. So we can only prevent the user from adding a annotation that belongs to an property with type Int to a property with type String at compile time.

In the future the goal IMHO should be to make those checks inside the idea plugin to provide better IDE support.

Resolves #69 .

### DIFF
--- a/godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
+++ b/godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
@@ -77,20 +77,6 @@ annotation class Flags
  */
 @Target(AnnotationTarget.PROPERTY)
 @Retention(AnnotationRetention.RUNTIME)
-annotation class Layers2DPhysics
-
-/**
- * Can only be used on Int properties!
- */
-@Target(AnnotationTarget.PROPERTY)
-@Retention(AnnotationRetention.RUNTIME)
-annotation class Layers3DPhysics
-
-/**
- * Can only be used on Int properties!
- */
-@Target(AnnotationTarget.PROPERTY)
-@Retention(AnnotationRetention.RUNTIME)
 annotation class Layers2DRender
 
 /**
@@ -98,7 +84,21 @@ annotation class Layers2DRender
  */
 @Target(AnnotationTarget.PROPERTY)
 @Retention(AnnotationRetention.RUNTIME)
+annotation class Layers2DPhysics
+
+/**
+ * Can only be used on Int properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
 annotation class Layers3DRender
+
+/**
+ * Can only be used on Int properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Layers3DPhysics
 
 /**
  * Can only be used on String and File properties!
--- a/godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
+++ b/godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
@@ -10,6 +10,34 @@ We should move that to the idea plugin at some point in time to provide IDE help
  */
 
 /**
+ * Can only be used on Int properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class IntRange(val start: Int, val end: Int, val step: Int = -1, val or: Range = Range.NONE)
+
+/**
+ * Can only be used on Float properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class FloatRange(val start: Float, val end: Float, val step: Float = -1f, val or: Range = Range.NONE)
+
+/**
+ * Can only be used on Double properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class DoubleRange(val start: Double, val end: Double, val step: Double = -1.0, val or: Range = Range.NONE)
+
+/**
+ * Can only be used on Float and Double properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ExpRange(val start: Float, val end: Float, val step: Float = -1f)
+
+/**
  * This property hint is optional!
  *
  * A Enum that is visible to the editor is always marked as PROPERTY_HINT_ENUM if not defined otherwise
@@ -22,6 +50,20 @@ We should move that to the idea plugin at some point in time to provide IDE help
 @Target(AnnotationTarget.PROPERTY)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class EnumTypeHint(val enumerateAsInt: Boolean = false)
+
+/**
+ * Can only be used on Float and Double properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ExpEasing(val attenuation: Boolean = false, val inout: Boolean = true)
+
+/**
+ * Can only be used on Float and Double properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Lenght(val lenght: Int = -1)
 
 /**
  * Can only be used on Map<Enum, Boolean> properties!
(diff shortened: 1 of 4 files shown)


godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
```diff
@@ -0,0 +1,246 @@
+package godot.annotation
```
piiertho, line 1: Hi ! Great job !
It seems it misses those hint flags:
```
PROPERTY_HINT_LENGTH
PROPERTY_HINT_SPRITE_FRAME
PROPERTY_HINT_KEY_ACCEL
PROPERTY_HINT_IMAGE_COMPRESS_LOSSY
PROPERTY_HINT_IMAGE_COMPRESS_LOSSLESS
PROPERTY_HINT_OBJECT_TOO_BIG
PROPERTY_HINT_INT_IS_OBJECTID
PROPERTY_HINT_MAX
```
Property hint is implemented [here](https://github.com/godotengine/godot/blob/3.2/core/object.h).
Also shouldn't we have hints in same order than in the engine, it would be easier to maintain.

piiertho (changes requested):

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
```diff
@@ -0,0 +1,246 @@
...
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class SaveFile(vararg val extensions: String = [])
+
+/**
+ * Can only be used on String and File properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class GlobalDir
```
raniejade: maybe drop this in favor of `Dir` with `global` property, which defaults to `false`?

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+ * Can only be used on String and File properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class GlobalFile(vararg val extensions: String = [])
```
raniejade: same here, drop in favor of `File` with global property.

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
```diff
@@ -0,0 +1,246 @@
...
+ * A Enum that is visible to the editor is always marked as PROPERTY_HINT_ENUM if not defined otherwise
+ * through another PropertyTypeHintAnnotation. It then will enumerate using strings.
+ *
+ * You only have to use this annotation if you want to enumerate enums with ints in the editor.
+ *
+ * Can only be used on Enum properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class EnumTypeHint(val enumerateAsInt: Boolean = false)
```
raniejade, line 52: What benefit do we get in `enumerateAsInt`? Or is it for the matter of completeness?

raniejade (commented):
This looks great! It seems we implemented every (almost) hint godot supports, unfortunately not all hints are well document so behaviour is unknown to us. This is a crazy idea, WDYT of just implementing hints that we know the behaviour of? Not a fan of completeness if we know little about what the other hints do!

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
chippmann: Dammit... Completely overlooked that file. That would have made it soo much easier ^^
Thanks I'll maybe implement them (depending on the outcome of the discussion below.

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
chippmann: Yeah good idea. I was just focusing on parity with godot but i think your approach is the better one. Going fix it this evening.

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
chippmann: :+1:

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
chippmann: Only for completeness

chippmann:
> This looks great! It seems we implemented every (almost) hint godot supports, unfortunately not all hints are well document so behaviour is unknown to us. This is a crazy idea, WDYT of just implementing hints that we know the behaviour of? Not a fan of completeness if we know little about what the other hints do!

Yes I was thinking about this as well but took the other approach: just implement it and fix it if it fails.
What do you think @piiertho ?

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
chippmann: I hate it how even in code their documentation is either wrong or outdated if present at all...
`PROPERTY_HINT_RANGE, ///< hint_text = "min,max,step,slider; //slider is optional"`
there is no slider implemented, at the last param it either accepts `or_greater` or `or_lesser` -.-

chippmann: FYI: LENGHT is deprecated: https://github.com/godotengine/godot/blob/e2b66cacf78ae39b94df748e9740b98a1f011e77/doc/classes/%40GlobalScope.xml

chippmann: `PROPERTY_HINT_SPRITE_FRAME` and `PROPERTY_HINT_KEY_ACCEL` as well

raniejade:
I'm fine either way, let's see what @piiertho thinks.

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
```diff
@@ -0,0 +1,266 @@
...
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Flags
+
+/**
+ * Can only be used on Int properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Layers2DPhysics
```
piiertho: Not in same order as `object.h` in engine:
```cpp
	PROPERTY_HINT_LAYERS_2D_RENDER,
	PROPERTY_HINT_LAYERS_2D_PHYSICS,
	PROPERTY_HINT_LAYERS_3D_RENDER,
	PROPERTY_HINT_LAYERS_3D_PHYSICS,
```
I let you choose if you merge with that order or not.

Btw great work !

piiertho (approved):

godot-kotlin/godot-library/src/nativeMain/kotlin/godot/annotation/PropertyTypeHintAnnotation.kt
chippmann: oh not seen this sorry. I will cahnge it

piiertho:
> > This looks great! It seems we implemented every (almost) hint godot supports, unfortunately not all hints are well document so behaviour is unknown to us. This is a crazy idea, WDYT of just implementing hints that we know the behaviour of? Not a fan of completeness if we know little about what the other hints do!
> 
> Yes I was thinking about this as well but took the other approach: just implement it and fix it if it fails.
> What do you think @piiertho ?

I think that we should implement them and them fix if it fails. Because if we never know them, we never will implement them.

PR #82, opened 2020-03-30
Rework IntelliJ Setup and add documentation

raniejade:
This PR does multiple things but mainly on simplifying IntelliJ IDEA setup.
- Remove the previously tracked `*.iml` files.
- Make `samples/mini-games` as a normal gradle build (previously a composite build). This means that testing out the samples require publishing the project locally.

Instructions for importing into IntelliJ IDEA can be found on the root `README.md` file, while I also added some additional documentation for the samples (which can be found in `samples/README.md`).


samples/mini-games/settings.gradle.kts
```diff
@@ -1,15 +1,18 @@
-// TODO: remove this once we have publishing setup for all artifacts
-includeBuild("../../") {
-    dependencySubstitution {
-        substitute(module("com.utopia-rise:godot-gradle-plugin")).with(project(":godot-gradle-plugin"))
+// without this doing:
```
raniejade, line 1: this is sadly required so that we can use the `plugins` block.

settings.gradle.kts
```diff
@@ -12,6 +12,10 @@ subdir("entry-generator") {
     include("godot-entry-generator")
 }
 
+// include the samples as composite builds
+// so they get included in IntelliJ IDEA!
+includeBuild("samples/mini-games")
```
raniejade: so that we only need one IntelliJ project.

plugins/godot-gradle-plugin/build.gradle.kts
```diff
@@ -31,4 +31,28 @@ dependencies {
     implementation(kotlin("stdlib"))
     implementation(kotlin("gradle-plugin"))
     implementation(kotlin("gradle-plugin-api"))
+}
+
+tasks {
+    val sourceJar by creating(Jar::class) {
+        archiveBaseName.set("godot-gradle-plugin")
```
chippmann: Couldn't we use the project name here instead?

plugins/godot-gradle-plugin/build.gradle.kts
```diff
@@ -31,4 +31,28 @@ dependencies {
...
+    }
+}
+
+publishing {
+    publications {
+        // this is only used for publishing locally.
+        val godotPlugin by creating(MavenPublication::class) {
+            pom {
+                groupId = "com.utopia-rise"
+                artifactId = "godot-gradle-plugin"
```
chippmann: Same here

samples/mini-games/settings.gradle.kts
chippmann: I would also add a todo here to remove it like you have done below

chippmann (changes requested):

samples/mini-games/settings.gradle.kts
raniejade: yep, I'll add it to the samples README though. We still need some documentation on how to try a local build.

settings.gradle.kts
```diff
@@ -12,6 +12,15 @@ subdir("entry-generator") {
     include("godot-entry-generator")
 }
 
+// TODO: remove this once the samples are not using the local artifacts
+val ignoreSamples: String? by extra
```
raniejade, line 16: @chippmann added a temporary param here. Will be removed the samples are using the actual published artifacts. This is needed for the first time you import the project into IJ!. So before importing you need to run: `./gradlew publishToMavenLocal -PignoreSamples`.

.idea/misc.xml
```diff
@@ -4,5 +4,5 @@
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="11" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_11" default="false" project-jdk-name="11" project-jdk-type="JavaSDK" />
```
chippmann: It again changed to jdk 11. Shouldn't it be jdk 8 for compatibility?

plugins/godot-gradle-plugin/build.gradle.kts
```diff
@@ -31,4 +31,28 @@ dependencies {
...
+        from(sourceSets["main"].allSource)
+    }
+}
+
+publishing {
+    publications {
+        // this is only used for publishing locally.
+        val godotPlugin by creating(MavenPublication::class) {
+            pom {
+                groupId = "com.utopia-rise"
```
chippmann: Sorry could have seen this earlier...
Can't we define this globally somewhere? as it's the same for all publications in all modules?

settings.gradle.kts
chippmann: ok but you need to add it to the docs as well because there the command is still without the new temporary argument

samples/README.md
```diff
@@ -0,0 +1,28 @@
+# godot-kotlin samples
+The following will show you how to test out a local build of godot-kotlin.
+
+## Publish artifacts locally
+For the moment, trying out the samples requires publishing the project locally. In order to do so,
+just run the following command:
+```shell script
+./gradlew publishToMavenLocal
```
chippmann: new temp argument missing

chippmann (changes requested):

plugins/godot-gradle-plugin/build.gradle.kts
raniejade: I'll use `project.group` for now as it has the value set to `com.utopia-rise`. We can generalise this later once we have the other modules that needs publishing.

.idea/misc.xml
raniejade: Yeah, I'll probably just ignore this file now. I've set it 8 a couple of times now.

.idea/compiler.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
chippmann: Why dfo we need this file?

chippmann (changes requested):

.idea/compiler.xml
raniejade: removed, I missed it out!

chippmann (approved):

PR #84, opened 2020-03-30
Implement annotation-processor

### LINKED ISSUE #71: Implement godot-annotation-processor
Implement a annotation processor like here: https://github.com/utopia-rise/godot-kotlin/tree/master/tools/godot-annotation-processor

The annotation processor should listen to the annotation's defined here: #68 , #70 , #69.

It should gather the descriptor classes in these lists:
- classDescriptors
- propertyDescriptors
- propertyDescriptors (for signal annotations)
- funcDescriptors

It should gather those informations using the mpApt project: https://github.com/Foso/MpApt
But should NOT pass any MpApt specific informations to the godot-entry-generator so it does not need any dependency to MpApt!

chippmann:
This implements the annotation processor.

It cannot be tested at the moment as it does not function without the compiler plugin. But as not much has changed to the last annotation processor, I have no reason to believe it's not working.

Closes #71

### DIFF
--- a/entry-generation/godot-annotation-processor/src/main/kotlin/godot/annotation/processor/GodotAnnotationProcessor.kt
+++ b/entry-generation/godot-annotation-processor/src/main/kotlin/godot/annotation/processor/GodotAnnotationProcessor.kt
@@ -9,9 +9,10 @@ import org.jetbrains.kotlin.descriptors.FunctionDescriptor
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
 import org.jetbrains.kotlin.platform.TargetPlatform
 import java.lang.instrument.IllegalClassFormatException
-import java.nio.file.Paths
 
-class GodotAnnotationProcessor : AbstractProcessor() {
+class GodotAnnotationProcessor(
+    private val entryGenerationOutputDir: String
+) : AbstractProcessor() {
     override fun getSupportedAnnotationTypes(): Set<String> =
         setOf(
             "godot.annotation.RegisterClass",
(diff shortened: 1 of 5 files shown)


entry-generation/godot-annotation-processor/src/main/kotlin/godot/annotation/processor/GodotAnnotationProcessor.kt
```diff
@@ -0,0 +1,106 @@
...
+    override fun processingOver() {
+        //TODO: entry generation
+    }
+
+    /**
+     * Returns the path as a String to which the generated code should be written
+     * @return directory path to where the generated code should be written to as a String
+     */
+    private fun getKaptGeneratedDirectory(): String {
+        return "${Paths.get("").toAbsolutePath()}/build/godot/entries"
```
raniejade: This value can be passed from gradle to the compiler plugin via the subplugin, can we make this a property of `GodotAnnotationProcessor`?

raniejade (changes requested):
Looking good, just one comment!

entry-generation/godot-annotation-processor/src/main/kotlin/godot/annotation/processor/GodotAnnotationProcessor.kt
chippmann: Sure. Good idea.

piiertho (approved):
LGTM

raniejade (approved):

PR #87, opened 2020-04-01
Feature/api code generation as gradletask

### LINKED ISSUE #86: Add API code generation as gradle task

For the merge purpose, we decided to set api code generation as a gradle task.

Not appropriate

See @raniejade repository

Move code to `buildSrc` of project.
Add a task implementation.
Call task from `godot-library` gradle.

Not appropriate.

Not appropriate.

piiertho:
This resolves #86 
It add api code generation, code will not compile as it is missing core types.
Also the api code generation would be changed during merge.
It setups source sets for `godot-library`.


buildSrc/src/main/kotlin/godot/codegen/Argument.kt
```diff
@@ -0,0 +1,53 @@
+package godot.codegen
+
+import com.beust.klaxon.Json
+
+
+class Argument(
+        @Json(name = "name")
```
raniejade, line 7: indentation feels a bit odd, should this be 4 spaces instead of 8?

buildSrc/src/main/kotlin/godot/codegen/Class.kt
```diff
@@ -0,0 +1,265 @@
+package godot.codegen
+
+import com.beust.klaxon.Json
+import com.squareup.kotlinpoet.*
+import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import java.io.File
+
+
+class Class(
+        @Json(name = "name")
```
raniejade, line 10: same here (odd indentation)

raniejade (changes requested):
Looking, got a few comments and I think a few files in `.idea` is accidentally committed? `.idea/artifacts/*` for example.

buildSrc/build.gradle.kts
```diff
@@ -10,4 +10,8 @@ repositories {
 
 dependencies {
     implementation(kotlin("gradle-plugin", version = "1.3.71"))
+    implementation("com.beust:klaxon:3.0.1") //5.2
```
chippmann, line 13: This version is extremely old (you see the commented 5.2. I did this when we where taking over the project. This was the current version back then). I would at least update it to the current version. But personally i would rather use something more well known like Gson, Moshi or jackson

buildSrc/build.gradle.kts
```diff
@@ -10,4 +10,8 @@ repositories {
...
+
+    //Remember to change version according to DependenciesVersion
+    implementation("com.squareup:kotlinpoet:1.5.0")
```
chippmann, line 16: I would add the const variable in that file as well and add the comment there as well (even though it's not used at the moment. Just so the person that adds that const variable could forget that this exists or simply don't know.

buildSrc/src/main/kotlin/godot/codegen/Class.kt
chippmann: In this case the indentation comes from the fact that the closing bracket of the constructor is not on a new line. That was probably also the case on the one above but the file wasn't auto formatted afterwards.
@piiertho I suggest that you activate the checkbox "format code" inside the commit dialog in Intellij  so you don't have to remember to format each file :-)

godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -1,18 +1,35 @@
...
-//TODO: this needs to be properly configured! This is just a basic setup to be able to implement the annotations
 kotlin {
-    linuxX64()
-    mingwX64()
-    macosX64()
-
-    val nativeMain = sourceSets.create("nativeMain")
+    macosX64("macos")
+//    linuxX64("linux")
+//    mingwX64("windows")
```
chippmann, line 11: Targets are commented

godot-kotlin/godot-library/src/main/nativeInternal/kotlin/godot/internal/Utils.kt
```diff
@@ -0,0 +1,28 @@
...
+import godot.gdnative.godot_method_bind
+import godot.gdnative.godot_method_bind_get_method
+import godot.gdnative.godot_variant_as_object
+import kotlinx.cinterop.COpaquePointer
+import kotlinx.cinterop.CPointer
+import kotlinx.cinterop.cstr
+import godot.core.Variant
+
+
+internal fun getMB(cl: String, m: String): CPointer<godot_method_bind> {
```
chippmann, line 14: Please call it `getMethodBind` ^^ At first glance i always think "What? what MB? how many MB's? get 50 MB's of ram or what?" xD

chippmann: Also pleas write out the parameter names

godot-kotlin/godot-library/src/main/nativeInternal/kotlin/godot/internal/Utils.kt
```diff
@@ -0,0 +1,28 @@
...
+    return godot_method_bind_get_method(cl, m) ?: throw NotImplementedError("Cannot get method bind for $m in $cl")
+}
+
+
+internal fun getSingleton(cl: String, clOld: String): COpaquePointer {
```
chippmann, line 19: please write out the parameter names

godot-kotlin/godot-library/src/main/nativeInterop/cinterop/registrations.h
```diff
@@ -0,0 +1,216 @@
+#ifndef WRAPPER_REGISTRATIONS_H
```
chippmann, line 1: why do we still need this file?

godot-kotlin/godot-library/src/main/nativeInterop/cinterop/godot.def
```diff
@@ -0,0 +1,2 @@
+package = godot.gdnative
+headers = gdnative_api_struct.gen.h registrations.h
```
chippmann, line 2: same question: why do we still need the registrations.h file?

chippmann (changes requested):
I didn't look at the actual codeGen logic yet (I would still wait until it compiles with the actual merge).
But one question about it: am i blind or didn't you yet add the changes regarding inline classes that we talked about on discord?

godot-kotlin/godot-library/build.gradle.kts
piiertho: oops, missed this

piiertho:
I close this one, it will be seperated into two PR, one for source sets #88 and one for api code generation.

PR #91, opened 2020-04-10
Update mpapt to release version

raniejade:
I don't have the snapshot version locally so the builds fails for me. Release build is already available: https://github.com/Foso/MpApt/commit/005af4a6a37b44aa161789503ef4a767fd4855fa


chippmann (approved):
Yeah. Did not have time yesterday to Update it. Thx a lot.

PR #92, opened 2020-04-10
Implement the casting system

raniejade:
This adds a `TypeManager` object that is responsible for casting a native godot object to the appropriate kotlin type.

Firstly, in order for this to work we need to register all user defined (as soon as they are registered to Godot) and engine types (this can be generated). There are two register methods: `registerEngineType` and `registerUserType`, their name alone should be descriptive enough to what type it should be used against.

To properly cast a native godot object, please use `TypeManager.wrap(ptr)`. As an example, lets hypothetically implement the `getParent` method of `Node`.

```kotlin
fun getParent(): Node? {
	val parent = call_get_parent_method_bind(this.ptr)
	return parent?.let { TypeManager.wrap(it) }
}
```


buildSrc/src/main/kotlin/DependenciesVersions.kt
```diff
@@ -1,4 +1,4 @@
 object DependenciesVersions {
-    const val mpaptVersion: String = "0.8.5-SNAPSHOT"
+    const val mpaptVersion: String = "0.8.5"
```
raniejade, line 2: temporary until https://github.com/utopia-rise/godot-kotlin/pull/91 is merged

godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -10,19 +10,20 @@ kotlin {
     linuxX64("linux")
     mingwX64("windows")
 
-    val internalSourceSet = sourceSets.create("nativeInternal")
-    val coreSourceSet = sourceSets.create("nativeCore") { dependsOn(internalSourceSet) }
-    val generatedSourceSet = sourceSets.create("nativeGen") { dependsOn(coreSourceSet) }
-    val publicSourceSet = sourceSets.create("nativePublic") { dependsOn(generatedSourceSet) }
+    // val internalSourceSet = sourceSets.create("nativeInternal")
```
raniejade, line 13: so that cinterop will work (this will conflict with what @piiertho is working on)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -0,0 +1,87 @@
+package godot.core
```
raniejade, line 1: copy pasted directly from my binding.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/TypeManager.kt
```diff
@@ -0,0 +1,101 @@
+package godot.core
+
+import kotlinx.cinterop.*
+import kotlin.native.concurrent.AtomicReference
+import kotlin.native.concurrent.freeze
+
+// TODO: remove when the class generator is done.
+open class Object(__ignore: String?) {
```
raniejade, line 8: will remove once @piiertho is done with the engine types.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/TypeManager.kt
```diff
@@ -0,0 +1,101 @@
...
+    constructor(): this(null) {
+        // nada
+    }
+
+    internal lateinit var ptr: COpaquePointer
+
+    fun getClass(): String = ""
+}
+
+object ClassDB {
```
raniejade, line 19: same as above

.idea/jarRepositories.xml
```diff
@@ -16,5 +16,10 @@
       <option name="name" value="BintrayJCenter" />
       <option name="url" value="https://jcenter.bintray.com/" />
     </remote-repository>
+    <remote-repository>
```
chippmann: Should this file in general be Uploaded to the repo?

buildSrc/src/main/kotlin/DependenciesVersions.kt
raniejade: this shouldn't show up now, not sure why :shrug:

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -0,0 +1,87 @@
...
+        val extensions = checkNotNull(gdnative.pointed.extensions)
+        lateinit var nativescript: CPointer<godot_gdnative_ext_nativescript_api_struct>
+        (0 until extensionCount).forEach { i ->
+            val extension = checkNotNull(extensions[i])
+            val type = extension.pointed.type
+            when (GDNATIVE_API_TYPES.byValue(type)) {
+                GDNATIVE_API_TYPES.GDNATIVE_EXT_NATIVESCRIPT -> {
+                    nativescript = extension.reinterpret()
+                }
+                else -> {
```
chippmann, line 48: Shouldent we at least log something here?

gradle.properties
```diff
@@ -1 +1 @@
-kotlin.mpp.enableGranularSourceSetsMetadata=true
\ No newline at end of file
+#kotlin.mpp.enableGranularSourceSetsMetadata=true
```
chippmann, line 1: Why is it commented?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/TypeManager.kt
```diff
@@ -0,0 +1,102 @@
...
+    fun getParentClass(cls: String): String = ""
+}
+
+internal object TypeManager {
+    private val tags = AtomicReference(mutableListOf<COpaquePointer>().freeze())
+
+    /**
+     * Register a user defined type.
+     */
+    fun registerUserType(nativescriptHandle: COpaquePointer, className: String, factory: () -> Object) {
```
chippmann, line 29: Do we have to call this now from the entry file?

chippmann (changes requested):

.idea/jarRepositories.xml
raniejade: Not sure, but I guess we can ignore it. I think this one is generated based on gradle.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
raniejade: Nah, this case is totally okay since we only use a single gdnative extension. Otherwise, the log will always show when godot loads which doesn't affect the user and they can't do anything either.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/TypeManager.kt
raniejade: Yes, after we register a class to godot.

gradle.properties
raniejade: Not being used, I do want to keep it there though so that once kotlin fixes the cinterop issues for intermediate source sets we can just comment this out.

.idea/jarRepositories.xml
raniejade: removed.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
chippmann: :+1:

chippmann (approved):
Thanks a lot for the work! :-)

piiertho (approved):

PR #95, opened 2020-04-11
EntryGeneration: Classes and Functions

chippmann:
This is a first PR for the entry generation. It can register classes and functions.

It does not create any GDNS file yet.


entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/mapper/RpcModeAnnotationMapper.kt
```diff
@@ -0,0 +1,22 @@
...
+
+object RpcModeAnnotationMapper {
+    /**
+     * maps RpcModeAnnotations and RPCMode registration enums to ClassNames to provide import informations for kotlinPoet
+     */
+    fun mapRpcModeAnnotationToClassName(rpcModeAnnotationAsString: String): ClassName {
+        return when (rpcModeAnnotationAsString) {
+            "godot.registration.RPCMode.DISABLED" -> ClassName("godot.registration.RPCMode", "DISABLED")
+            "godot.registration.RPCMode.REMOTE" -> ClassName("godot.registration.RPCMode", "REMOTE")
+            "godot.registration.RPCMode.SYNC" -> ClassName("godot.registration.RPCMode", "SYNC")
```
piiertho: SYNC is deprecated.
I think deprecated should stay as they are not removed yet. But in this case, we miss SLAVE.
https://docs.godotengine.org/en/3.1/classes/class_multiplayerapi.html#enumerations

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/transformer/TypeDeclarationsToClassWithMemberTransformer.kt
```diff
@@ -0,0 +1,44 @@
+package godot.entrygenerator.transformer
+
+import godot.entrygenerator.model.ClassWithMembers
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+
+object TypeDeclarationsToClassWithMemberTransformer {
+
+    fun transform(
```
piiertho: I would rather prefer a top level method, but that's a personal preference.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/ClassHandle.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun registerFunction(methodName: String, methodRef: COpaquePointer, rpcMode: RPCMode) {
+        disposables.add(methodRef)
+        memScoped {
+            val attribs = cValue<godot_method_attributes> {
+                rpc_type = when (rpcMode) {
+                    RPCMode.DISABLED -> GODOT_METHOD_RPC_MODE_DISABLED
+                    RPCMode.REMOTE -> GODOT_METHOD_RPC_MODE_REMOTE
+                    RPCMode.SYNC -> GODOT_METHOD_RPC_MODE_SYNC
+                    RPCMode.MASTER -> GODOT_METHOD_RPC_MODE_MASTER
+                    RPCMode.SLAVE -> GODOT_METHOD_RPC_MODE_SLAVE
```
piiertho: Same here for RPC.

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/mapper/RpcModeAnnotationMapper.kt
chippmann: SLAVE was renamed to PUPPET for godot 3.1 i think

chippmann: and i think we can remove the deprecated ones

chippmann: i removed the deprecated one if it's ok for you

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -24,7 +26,12 @@ class GodotPlugin : Plugin<Project> {
 
     private fun setupExtensionDefaults(project: Project, godot: GodotExtension) {
         with(godot) {
-            platforms(*Platform.values())
+            // we don't have godot-library in the mobile targets yet, limit these to desktop for now
```
piiertho, line 29: We should add a comment in both build file of godot-library and here to remember to add mobile targets in both, I think.

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/transformer/TypeDeclarationsToClassWithMemberTransformer.kt
chippmann: changed it

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
chippmann: added

piiertho (changes requested):
Few comment otherwise I'm ok. Really smart :)

piiertho (approved):
Ok for merging it ! :)

raniejade (approved):

PR #129, opened 2020-04-14
Implement property registration entry generation

### LINKED ISSUE #97: EntryGenerator DefaultValues (1): Support for primitive Types & Enums
Implement default value support for primitiveTypes (expect array which are considered primitive in this case as well).

This includes the following:
- Int
- Long
- Float
- Double
- String
- Boolean
- Enums

```kotlin
@RegisterProperty
var int = 100

@RegisterProperty
var long = 100L

@RegisterProperty
var float = 1.0f

@RegisterProperty
var double = 2.0

@RegisterProperty
var str = "hello"

@RegisterProperty
var bool = true

@RegisterProperty
var enum = TestEnum.ENUM1
```

**Implementation hints:**
```kotlin
```

raniejade:
Resolves #97. It only supports primitives and enums at the moment.


entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/PropertyDefaultValueProvider.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.jetbrains.kotlin.js.resolve.diagnostics.findPsi
+import org.jetbrains.kotlin.psi.*
+import org.jetbrains.kotlin.psi.psiUtil.referenceExpression
+import org.jetbrains.kotlin.resolve.BindingContext
+import org.jetbrains.kotlin.resolve.bindingContextUtil.getReferenceTargets
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+import org.jetbrains.kotlin.resolve.source.KotlinSourceElement
+import org.jetbrains.kotlin.serialization.deserialization.descriptors.DeserializedClassConstructorDescriptor
+
+object PropertyDefaultValueProvider {
```
chippmann: This is not required anymore (I left it after the refactoring proposal because we might needed it but after your changes in `RegistrationValuesHandler` they its not needed anymore)

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/EnumRegistrationValuesHandler.kt
```diff
@@ -0,0 +1,38 @@
...
+    override fun getDefaultValueStringTemplateValues(): Array<Any> {
+        return if (propertyDescriptor.isLateInit) {
+            arrayOf("null")
+        } else {
+            arrayOf(ClassName("godot.core", "Variant"), getDefaultValue())
+        }
+    }
+
+    override fun getPropertyTypeHint(): ClassName {
+        throw UnsupportedOperationException("Hint type for enums is handled elsewhere!")
```
chippmann: Maybe we should also say where

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/EnumRegistrationValuesHandler.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    override fun getHintString(): String {
+        throw UnsupportedOperationException("Hint string for enums is handled elsewhere!")
```
chippmann: same here

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+abstract class RegistrationValuesHandler(
+    val propertyDescriptor: PropertyDescriptor,
+    val bindingContext: BindingContext
+) {
+    abstract fun getDefaultValueStringTemplate(): String
+    abstract fun getDefaultValueStringTemplateValues(): Array<Any>
+    abstract fun getPropertyTypeHint(): ClassName
+    abstract fun getHintString(): String
+
+    protected fun getDefaultValue(): String {
```
chippmann: Not sure if we really should have this method
the idea of `getDefaultValueStringTemplate` and `getDefaultValueStringTemplateValues` was that you could generate the template and the corresponding values to let kotlinPoet handle all imports also the ones inside the `Variant(...)`.

godot-kotlin/godot-library/src/nativePublic/kotlin/godot/annotation/DefaultValueAnnotation.kt
```diff
@@ -0,0 +1,10 @@
+package godot.annotation
+
+interface Defaults {
```
chippmann: Note to self: This is not needed (sorry forgot to remove it)

chippmann (changes requested):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/DefaultValueHandlerProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+        return when  {
+            KotlinBuiltIns.isInt(propertyDescriptor.type)
+                || KotlinBuiltIns.isInt(propertyDescriptor.type)
+                || KotlinBuiltIns.isLong(propertyDescriptor.type)
+                || KotlinBuiltIns.isFloat(propertyDescriptor.type)
+                || KotlinBuiltIns.isDouble(propertyDescriptor.type)
+                || KotlinBuiltIns.isBoolean(propertyDescriptor.type)
+                || KotlinBuiltIns.isString(propertyDescriptor.type) -> PrimitiveRegistrationValuesHandler(propertyDescriptor, bindingContext)
+            propertyDescriptor.type.isEnum() -> EnumRegistrationValuesHandler(propertyDescriptor, bindingContext)
+            else -> TODO()
```
piiertho, line 22: Maybe add a reason in todo ? like "Only primitives can be registered for now" ?

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/mapper/PropertyHintTypeMapper.kt
```diff
@@ -0,0 +1,298 @@
...
+            "godot.annotation.FloatRange" ->
+                getRangeTypeHint(annotationDescriptor, arrayOf(Float::class), propertyDescriptor)
+            "godot.annotation.DoubleRange" ->
+                getRangeTypeHint(annotationDescriptor, arrayOf(Double::class), propertyDescriptor)
+            "godot.annotation.ExpRange" ->
+                getRangeTypeHint(annotationDescriptor, arrayOf(Float::class, Double::class), propertyDescriptor)
+            "godot.annotation.EnumTypeHint" -> getEnumTypeHint(propertyDescriptor)
+            "godot.annotation.ExpEasing" -> getExpEasingTypeHint(annotationDescriptor, propertyDescriptor)
+            "godot.annotation.Lenght" -> throw NotImplementedError("This annotation is not yet implemented") //getLengthTypeHint(annotationDescriptor, propertyDescriptor)
+            "godot.annotation.Flags" -> throw NotImplementedError("This annotation is not yet implemented")//getFlagsTypeHint(propertyDescriptor)
```
piiertho: Maybe we can add the name of the annotation in Exception message ?

godot-kotlin/godot-library/src/nativePublic/kotlin/godot/annotation/RegisterAnnotation.kt
```diff
@@ -14,7 +14,7 @@ annotation class RegisterClass(val isTool: Boolean = false)
 @Target(AnnotationTarget.PROPERTY)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class RegisterProperty(
-    val visibleInEditor: Boolean = false,
+    val visibleInEditor: Boolean = true,
```
piiertho, line 17: Why default value is true ?

piiertho (changes requested):
Few comments, but that's awesome work !

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/DefaultValueHandlerProvider.kt
chippmann: I don't think this is necessary bacause we will have implemented all supported before the first beta release

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/mapper/PropertyHintTypeMapper.kt
chippmann: :+1:

godot-kotlin/godot-library/src/nativePublic/kotlin/godot/annotation/RegisterAnnotation.kt
chippmann: Most properties you want to register you also want to be visible in the editor.
You only don't want it to be visible for accessign through gdScript. here we took a kotlin first approach. We were expecting most users who use this binding to just use kotlin and to use other languages lesser. so most of the times the user would have to provide a boolean.

chippmann: I think it's mainly a question how we want to present the binding. as a additional way to programming for godot. or the main way for programming.
Personally I'm for the last option

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
raniejade: I think this should be common with all supported cases, the reason why I added it as a common method in the base class is because of constructor calls. Imagine, `Vector2(1.0, 2.0)`, the check happening is recursive. Firstly it will go thru the check `expression is KtCallExpression`, then for each parameter recursively call `getDefaultValueExpression`. If we don't have this then we have de-duplicate all the checks in all registration handles. 

Ideally this function will transform `SomeCoreClass(SomeCoreClass.ENUM.VALUE, 2.0)` to `godot.core.SomeCoreClass(godot.core.SomeCoreClass.ENUM.VALUE, 2.0)`. I just need to make that change.

godot-kotlin/godot-library/src/nativePublic/kotlin/godot/annotation/RegisterAnnotation.kt
raniejade: I think most users will use this binding to create games, so `visibleInEditor = true`.

piiertho: I agree with you, if performances permit it :)
So LGTM.

piiertho (approved):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
chippmann: I see your point but then ideally this should return a `Pair<String, Any>` with the string template as first and the values as `ClassName` or literals in second. Then KotlinPoet can handle imports. I would much more like that

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
```diff
@@ -0,0 +1,97 @@
...
+import org.jetbrains.kotlin.serialization.deserialization.descriptors.DeserializedClassConstructorDescriptor
+
+abstract class RegistrationValuesHandler(
+    val propertyDescriptor: PropertyDescriptor,
+    val bindingContext: BindingContext
+) {
+    abstract fun getPropertyTypeHint(): ClassName
+    abstract fun getHintString(): String
+
+    fun getDefaultValue(): Pair<String, Array<Any>> {
```
raniejade, line 22: @chippmann

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
```diff
@@ -0,0 +1,97 @@
...
+        params.add(ClassName("godot.core", "Variant"))
+        params.addAll(defaultValue.second)
+        return "%T(${defaultValue.first})" to params.toTypedArray()
+    }
+
+    private fun getDefaultValueExpression(expression: KtExpression): Pair<String, Array<Any>>? {
+        if (expression is KtConstantExpression) {
+            return return "%L" to arrayOf(expression.text)
+        } else if (expression is KtStringTemplateExpression && !expression.hasInterpolation()) {
+            return return "%S" to arrayOf(expression.text)
```
chippmann: Have you tested this? I think this will result in `""text""`

chippmann (changes requested):

chippmann (approved):

PR #131, opened 2020-04-14
Refactor hintType and hintString generation and implement coreType re…

### LINKED ISSUE #98: EntryGenerator DefaultValues (2): Support for core types
Implement default value support for coreTypes (except array 's and resources!).

```kotlin
@RegisterProperty
var nodePath: NodePath = NodePath("")
```

**Implementation hints:**
```kotlin
```

chippmann:
this resolves #98 and enables default value generation and type hint generation for core types.

It also refactors the said generation to happen in the individual `RegistrationValueHandler`s rather than in one big `PropertyHintMapper` (file still exists thoug for reference until all `RegistrationValueHandler`s are implemented)

### DIFF
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/exceptions/WrongAnnotationUsageException.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/exceptions/WrongAnnotationUsageException.kt
@@ -1,0 +1,14 @@
+package godot.entrygenerator.exceptions
+
+import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+class WrongAnnotationUsageException(
+    propertyDescriptor: PropertyDescriptor,
+    propertyHintAnnotation: AnnotationDescriptor?,
+    effectiveType: String? = null
+) : Exception(
+    "You annotated ${propertyDescriptor.fqNameSafe} with @${propertyHintAnnotation?.fqName?.asString()?.split(".")
+        ?.last()} which ${if (effectiveType != null) "is only applicable to properties of type $effectiveType" else "cannot be applied on properties of type ${propertyDescriptor.type}"}"
+)
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
@@ -1,0 +1,37 @@
+package godot.entrygenerator.extension
+
+import org.jetbrains.kotlin.types.KotlinType
+
+
+fun KotlinType.isCoreType(): Boolean {
+    return coreTypes.contains(this.toString())
+}
+
+private val coreTypes = listOf(
+    "GDArray",
+    "Basis",
+    "Color",
+    "Dictionary",
+    "GodotError",
+    "NodePath",
+    "Plane",
+    "PoolByteArray",
+    "PoolIntArray",
+    "PoolRealArray",
+    "PoolStringArray",
+    "PoolVector2Array",
+    "PoolVector3Array",
+    "PoolColorArray",
+    "PoolIntArray",
+    "PoolRealArray",
+    "Quat",
+    "Rect2",
+    "AABB",
+    "RID",
+    "String",
+    "Transform",
+    "Transform2D",
+    "Variant",
+    "Vector2",
+    "Vector3"
+)
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/CoreTypeRegistrationValuesHandler.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/CoreTypeRegistrationValuesHandler.kt
@@ -1,0 +1,47 @@
+package godot.entrygenerator.generator.provider
+
+import com.squareup.kotlinpoet.ClassName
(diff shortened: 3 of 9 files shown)


entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/exceptions/WrongAnnotationUsageException.kt
```diff
@@ -0,0 +1,14 @@
+package godot.entrygenerator.exceptions
+
+import org.jetbrains.kotlin.descriptors.PropertyDescriptor
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+class WrongAnnotationUsageException(
```
raniejade, line 7: I guess this will be temporary until we moved to using `MessageCollector`?

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/CoreTypeRegistrationValuesHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+
+class CoreTypeRegistrationValuesHandler(
+    propertyDescriptor: PropertyDescriptor,
+    bindingContext: BindingContext
+) : RegistrationValuesHandler(propertyDescriptor, bindingContext) {
+
+    override fun getPropertyTypeHint(): ClassName {
+
+        return when (propertyHintAnnotation?.fqName?.asString()) {
+            "godot.annotation.ColorNoAlpha" ->
```
raniejade, line 16: unrelated but I think the annotation should be `@Color(noAlpha = true)`.

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
```diff
@@ -16,9 +20,26 @@ abstract class RegistrationValuesHandler(
     val propertyDescriptor: PropertyDescriptor,
     val bindingContext: BindingContext
 ) {
+    internal val propertyHintAnnotation = propertyDescriptor.getPropertyHintAnnotation()
+
     abstract fun getPropertyTypeHint(): ClassName
     abstract fun getHintString(): String
 
+    init {
+        checkHintAnnotationUsage()
```
raniejade, line 29: Something for later, but we need to check that the annotation hints used for a property makes sense.

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/exceptions/WrongAnnotationUsageException.kt
chippmann: Yes. But something like this should also be used for the message collector so the error messages are the same.

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/CoreTypeRegistrationValuesHandler.kt
chippmann: I dont think so. Only this typehint exists. In the other case you just export a color with no typehint. No need for a typehint annotation. Same as in gdScript

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
chippmann: Wdym?

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
```diff
@@ -0,0 +1,37 @@
+package godot.entrygenerator.extension
+
+import org.jetbrains.kotlin.types.KotlinType
+
+
+fun KotlinType.isCoreType(): Boolean {
```
piiertho, line 6: Have pretty much same methods in api code generation, maybe we can do common helper module for thoses later on.

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
piiertho: I think @raniejade means we should not be able to annotate with type hint annotations on properties that cannot have this type hint.

piiertho (approved):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/RegistrationValuesHandler.kt
chippmann: That's already handled through the `getPropertyTypeHint` and the`getHintString` methods in each `RegistrationValueHandler`

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
chippmann: Good idea

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/CoreTypeRegistrationValuesHandler.kt
raniejade: Ahh, that makes sense!

raniejade (approved):

PR #132, opened 2020-04-15
EntryRegistration: Implement GodotResource registration

### LINKED ISSUE #99: EntryGenerator DefaultValues (3): Support for Resources
Implement default value support for Resources and custom Resources.

```kotlin
@RegisterProperty
lateinit var resource: Resource //if not lateinit -> fail, customResources: just have to derive from Resource but are exported as Resource and not as custom type
```

**Implementation hints:**
- Resource must be defined as lateinit and does NOT support default values! -> this must fail otherwise at compile time!
- A custom resource should be registered as just a resource and be casted back to the custom type when set by godot
```kotlin
```

chippmann:
This closes #99 and fixes the registration of properties which are not visible in the editor by not providing a default value for them.

### DIFF
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
@@ -1,5 +1,6 @@
 package godot.entrygenerator.extension
 
+import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
 import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.typeUtil.supertypes
 
@@ -9,10 +10,11 @@ fun KotlinType.isCoreType(): Boolean {
 }
 
 fun KotlinType.isResource(): Boolean {
-    return this.toString() == "GodotResource"
+    return this.getJetTypeFqName(false) == "godot.Resource"
         || this
         .supertypes()
-        .any { it.toString() == "GodotResource" }
+        .map { it.getJetTypeFqName(false) }
+        .any { it == "godot.Resource" }
 }
 
 private val coreTypes = listOf(
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
@@ -1,10 +1,18 @@
 package godot.entrygenerator.extension
 
 import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.supertypes
 
 
 fun KotlinType.isCoreType(): Boolean {
     return coreTypes.contains(this.toString())
+}
+
+fun KotlinType.isResource(): Boolean {
+    return this.toString() == "GodotResource"
+        || this
+        .supertypes()
+        .any { it.toString() == "GodotResource" }
 }
 
 private val coreTypes = listOf(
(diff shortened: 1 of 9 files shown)


entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
```diff
@@ -1,12 +1,20 @@
...
 import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.supertypes
 
 
 fun KotlinType.isCoreType(): Boolean {
     return coreTypes.contains(this.toString())
 }
 
+fun KotlinType.isResource(): Boolean {
+    return this.toString() == "GodotResource"
```
raniejade: are we able to check the fully qualified name? Also shouldn't this be `Resource`?

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/ResourceRegistrationValuesHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+        return when (propertyHintAnnotation?.fqName?.asString()) {
+            //TODO: implement ResourceType
+            null -> ClassName("godot.gdnative.godot_property_hint", "GODOT_PROPERTY_HINT_NONE")
+            else -> throw WrongAnnotationUsageException(propertyDescriptor, propertyHintAnnotation)
+        }
+    }
+
+    override fun getHintString(): String {
+        //TODO: implement ResourceType
+        return ""
```
raniejade, line 31: The hint string is basically the class name of the type right?

raniejade (commented):
A few comments :)

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
```diff
@@ -1,6 +1,7 @@
 package example
 
 import godot.annotation.*
+import godot.core.GodotResource
```
piiertho: Resource is generated and so is in godot package.

piiertho (changes requested):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
chippmann: with this yes `kotlinType.getJetTypeFqName(false)` but maybe there is a better solution?

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/ResourceRegistrationValuesHandler.kt
chippmann: not that I'm aware of. I didn't find any information about this in the source code. I guess it's always empty for resources?

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
chippmann: ok i move the dummyClass

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/ResourceRegistrationValuesHandler.kt
chippmann: for the resource type it's not empty and it's the class name yes but that hint type annotation is not yet implemented

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
chippmann: currently it's named `GodotResource` @piiertho whats the final name?

piiertho (approved):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
piiertho: It is named Resource :)

chippmann: I renamed it :-)

PR #134, opened 2020-04-22
Only support VariantArray for property registration

raniejade:
As discussed this drops support for kotlin collection types in property registration. I also implemented `VariantArray<T>` in this PR :)

### DIFF
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
@@ -1,6 +1,5 @@
 package godot.entrygenerator.extension
 
-import org.jetbrains.kotlin.builtins.KotlinBuiltIns
 import org.jetbrains.kotlin.js.descriptorUtils.getJetTypeFqName
 import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.typeUtil.supertypes
@@ -18,19 +17,8 @@ fun KotlinType.isResource(): Boolean {
         .any { it == "godot.Resource" }
 }
 
-fun KotlinType.isVariantArray(): Boolean {
-    return this.getJetTypeFqName(false) == "godot.core.VariantArray"
-        || this
-        .supertypes()
-        .map { it.getJetTypeFqName(false) }
-        .any { it == "godot.core.VariantArray" }
-}
-
 fun KotlinType.isCompatibleList(): Boolean {
     return when {
-        KotlinBuiltIns.isListOrNullableList(this)
-            || KotlinBuiltIns.isArray(this)
-            || KotlinBuiltIns.isPrimitiveArray(this) -> true
         getJetTypeFqName(false) == "godot.core.VariantArray" -> true
         else -> false
     }
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/VariantArray.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/VariantArray.kt
@@ -1,33 +1,71 @@
 package godot.core
 
 import godot.gdnative.godot_array
-import kotlinx.cinterop.CValue
+import kotlinx.cinterop.*
 
-class VariantArray<T>(val value: CValue<godot_array>) : AbstractMutableList<T>() {
+class VariantArray<T> internal constructor(private val handle: CValue<godot_array>) : AbstractMutableList<T>() {
     constructor(): this(new())
 
     override fun add(index: Int, element: T) {
-        TODO("Not yet implemented")
+        memScoped {
+            checkNotNull(Godot.gdnative.godot_array_insert)(
+                handle.ptr,
+                index,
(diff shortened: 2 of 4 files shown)


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/VariantArray.kt
```diff
@@ -1,33 +1,71 @@
 package godot.core
 
 import godot.gdnative.godot_array
-import kotlinx.cinterop.CValue
+import kotlinx.cinterop.*
 
-class VariantArray<T>(val value: CValue<godot_array>) : AbstractMutableList<T>() {
+class VariantArray<T> internal constructor(private val handle: CValue<godot_array>) : AbstractMutableList<T>() {
```
piiertho: I need handle to be accessible for api code generation.
See here: https://github.com/utopia-rise/godot-kotlin/blob/34d27b0532360c120424de5e885cad235a4b1d4f/buildSrc/src/main/kotlin/godot/codegen/ICall.kt#L103

piiertho (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/VariantArray.kt
raniejade: fixed in https://github.com/utopia-rise/godot-kotlin/pull/134/commits/2d61abf7b70d4738995c57343f0dffa617f5e463

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
```diff
@@ -18,19 +17,8 @@ fun KotlinType.isResource(): Boolean {
         .any { it == "godot.Resource" }
 }
 
-fun KotlinType.isVariantArray(): Boolean {
-    return this.getJetTypeFqName(false) == "godot.core.VariantArray"
-        || this
-        .supertypes()
-        .map { it.getJetTypeFqName(false) }
-        .any { it == "godot.core.VariantArray" }
-}
```
chippmann, line 27: Why don't we support it anymore to extend the variant array class with your own implementation?

chippmann (changes requested):

piiertho (approved):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
raniejade: Hmm, why would you do that?

chippmann: provide additional functionality to variantArray. Setup some kind of type hierarchy with it and do stuff depending on the type of it. I don't know. There probably is a use case and i don't see why we can't or shouldn't support it now.

chippmann: for example implement functions in a custom variant array.

Custom VariantArray class Foo has function splitByType()
Custom VariantArray class Bar has function splitByName()

This can't be done through extension functions as one function should be only applicable to a specific instance of a custom variantArray

raniejade: I'm not sure that we can support that as `Variant` needs to be aware of that custom type. 

```kotlin
@RegisterProperty
var myCustomArray = MyCustomArray()
```

So when godot sets this property, it passes on a `Variant` which we call `unwrap` on it will return a `VariantArray`. In order to support that we need give that information to variant at runtime.

Also supporting it also means users can add additional properties to it: `class MyCustomArray(val data: ...): VariantArray<Int>` which are additional information we need to keep track at runtime.

I'm not convince the user value it adds outweighs the complexity needed to support it.

Extension function is IMHO a more idiomatic way to implement this use case (https://medium.com/@elizarov/extension-oriented-design-13f4f27deaee)

```kotlin
fun VariantArray<SomeType>.doSomethingWithSomeType() = ...
fun VariantArray<Int>.sum() = ...
```

raniejade: Also, I think we should really think of core types as final classes. They can't be inherited from GDScript as they are akin to primitives.

chippmann: Convinced

chippmann: @raniejade do you look with @piiertho regarding finalizing the core classes? I find that a good idea :-)

raniejade:
@chippmann are you happy with the PR? If so can you approve :pray:

chippmann (approved):
Sorry forgot to change the approval.

PR #137, opened 2020-04-24
GDNS file generation

chippmann:
This implements the gdns file generation and deletion.

### DIFF
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
@@ -24,7 +24,9 @@ object GdnsGenerator {
             val name = it.name.asString()
             val fqName = it.fqNameSafe.asString()
 
-            File(getGdnsFilePath(outputPath, fqName)).writeText(getGdnsFileContent(gdnLibFile, name, fqName))
+            File(getGdnsFilePath(outputPath, fqName).replaceAfterLast("/", "")).mkdirs()
+            File(getGdnsFilePath(outputPath, fqName))
+                .writeText(getGdnsFileContent(gdnLibFile, name, fqName))
         }
     }
 
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
@@ -6,7 +6,12 @@ import java.io.File
 
 object GdnsGenerator {
 
-    fun generateGdnsFiles(outputPath: String, gdnLibFile: String, classes: Set<ClassDescriptor>) {
+    fun generateGdnsFiles(
+        outputPath: String,
+        gdnLibFile: String,
+        cleanGeneratedGdnsFiles: Boolean,
+        classes: Set<ClassDescriptor>
+    ) {
         val existingGdnsFiles = File(outputPath)
             .listFiles()
             ?.filter { it.extension == "gdns" }
@@ -16,8 +21,10 @@ object GdnsGenerator {
         val obsoleteGdnsFiles = existingGdnsFiles
             .filter { !classNames.contains(it.name) }
 
-        obsoleteGdnsFiles.forEach {
-            it.delete()
+        if (cleanGeneratedGdnsFiles) {
+            obsoleteGdnsFiles.forEach {
+                it.delete()
+            }
         }
 
         classes.forEach {
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
@@ -28,23 +28,22 @@ object GdnsGenerator {
         }
 
         classes.forEach {
-            val name = it.name.asString()
             val fqName = it.fqNameSafe.asString()
 
             File(getGdnsFilePath(outputPath, fqName).replaceAfterLast("/", "")).mkdirs()
             File(getGdnsFilePath(outputPath, fqName))
-                .writeText(getGdnsFileContent(gdnLibFile, name, fqName))
+                .writeText(getGdnsFileContent(gdnLibFile, fqName))
         }
     }
 
-    private fun getGdnsFileContent(gdnLibFile: String, className: String, classFqName: String): String {
+    private fun getGdnsFileContent(gdnLibFile: String, classFqName: String): String {
         return """
             [gd_resource type="NativeScript" load_steps=2 format=2]
 
             [ext_resource path="res://$gdnLibFile" type="GDNativeLibrary" id=1]
 
             [resource]
-            resource_name = "$className"
+            resource_name = "$classFqName"
             class_name = "$classFqName"
             library = ExtResource( 1 )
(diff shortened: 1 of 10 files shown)


plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -82,13 +81,13 @@ class GodotPlugin : Plugin<Project> {
...
-                }
-                buildTask.dependsOn(cleanGeneratedFilesTask)
-            }
+//            if (godot.cleanupGeneratedFiles.get()) {
+//                val cleanGeneratedFilesTask = project.tasks.register<Delete>("cleanGeneratedFiles") {
+//                    group = GODOT_TASK_GROUP
+//                    delete(godot.gdnsDir.get())
+//                }
+//                buildTask.dependsOn(cleanGeneratedFilesTask)
+//            }
```
chippmann, line 91: Commented as discussed on discord

piiertho (approved):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
```diff
@@ -0,0 +1,56 @@
...
+
+            [resource]
+            resource_name = "$className"
+            class_name = "$classFqName"
+            library = ExtResource( 1 )
+        """.trimIndent()
+    }
+
+    private fun getGdnsFilePath(outputPath: String, classFqName: String): String {
+        return "$outputPath/${classFqName.replace(".", "/")}.gdns"
```
raniejade, line 53: Should we include the package name here? Godot has no concept of namespaces/packages, so including the package in the file path gives the illusion that it does. 

So if you have two classes with the same simple name but in different packages they will have separate gdns files but essentially points to a single class in Godot. I'm pretty sure Godot will fail if you register the same class name twice.

Now I'm thinking, should we include the fully qualified name of the class during registration? The convert `.` to a valid identifier character, maybe `__` (note double underscore)? `com.example.MyClass` will be registered to Godot as `com__example__MyClass`. With this we can keep the package name included in the file path for the gdns file. Note that the user will most probably don't have to deal with the fully qualified class name in Godot (i.e `com__example__MyClass`) as the way you attach scripts is via the gdns file.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
raniejade: Add a comment why? So we don't forget :)

raniejade (changes requested):

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
chippmann: :+1:

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/GdnsGenerator.kt
chippmann: To be honest i didn't thought of that. I simply used the package path as a directory structure so the gdns files are structured as well. Most of the time one defines his package names structurally sane so also the generated gdns file should follow the code structure IMO.
Regarding class names; Thats a good idea. As said i would only apply it for the registered class name, not for the path where the gdns file lies though.

So gdns:
```
[gd_resource type="NativeScript" load_steps=2 format=2]

[ext_resource path="res://mini-games.gdnlib" type="GDNativeLibrary" id=1]

[resource]
resource_name = "example__TestingClass"
class_name = "example.TestingClass"
library = ExtResource( 1 )
```

Registered ClassName: `example__TestClass`

But are there really no use cases where a user might need a class name?

chippmann: How c# does that btw?

chippmann: or the better question: how is it handled in other bindings like cpp?

raniejade: I think with C++ they just used the non-qualified name of the class. Not sure about C# though, how do you attach a C# script in the editor?

chippmann: It seems we can create resources with the same name (gdns files are just resource files in the end).  So yes we will just have to register a different name and then it should work :thinking: :
```
[gd_resource type="NativeScript" load_steps=2 format=2]

[ext_resource path="res://mini-games.gdnlib" type="GDNativeLibrary" id=1]

[resource]
resource_name = "TestingClass"
class_name = "example.TestingClass"
library = ExtResource( 1 )
```

raniejade: Also, what does `class_name = "example.TestingClass"` do? Is `.` a valid identifier?

chippmann: > I think with C++ they just used the non-qualified name of the class. Not sure about C# though, how do you attach a C# script in the editor?

Like with GDScript

chippmann: > Also, what does `class_name = "example.TestingClass"` do? Is `.` a valid identifier?

It points to the class in our lib. I guess. as we generated it like this in our old binding and it worked ^^

chippmann: This is one of our old generated ones:
```
[gd_resource type="NativeScript" load_steps=2 format=2]

[ext_resource path="res://samples.gdnlib" type="GDNativeLibrary" id=1]

[resource]
resource_name = "Main"
class_name = "godot.samples.games.pong.Main"
library = ExtResource( 1 )
```

raniejade: I'm also keen to register the normalized (`.` -> `__`) fully qualified name as the class name in Godot. 

GDNS should also mirror that class name, IMHO.
```
[resource]
resource_name = "example__TestingClass"
class_name = "example__TestingClass"
library = ExtResource( 1 )
```

As I mentioned before I don't think the user will never directly refer to the class by name in the editor but instead via the GDNS file. 

@piertho @chippmann WDYT?

chippmann: Worth a thought but how do we name the files? Just SimpleClassName as before or also fullay qualified?

raniejade: Same as before, but with the package reflected in the directory structure as default :)

chippmann: For reference: the discussion was continued on discord.

raniejade (approved):

PR #139, opened 2020-04-25
Implement `ResourceType` annotation

chippmann:


piiertho (approved):

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
```diff
@@ -76,6 +76,7 @@ class TestingClass : Object() {
     @RegisterProperty
     var variantTest = Variant("test")
 
+    @ResourceType(inherits = [Resource::class])
     @RegisterProperty
     lateinit var resourceVisibleInEditor: Resource
```
raniejade: Do we need this annotation? Can we just derived it from the type of the property? In GDScript it is done this way due to it not having type annotations:

```gdscript
export(Texture) var character_face
export(PackedScene) var scene_file
```

raniejade (changes requested):

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
chippmann: good point
I'll change it

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
```diff
@@ -52,3 +62,71 @@ private val coreTypes = listOf(
     "Vector2",
     "Vector3"
 )
+
+/**
+ * taken from the "inherited by" section here: https://docs.godotengine.org/en/stable/classes/class_resource.html
+ */
+private val resourceTypes = listOf(
+    "godot.Animation",
```
raniejade: I don't think we need to list all resource types, a check that a type inherits from `godot.Resource` should be enough. This also means that when new resource types are added, we don't have to update this list, it just works out of the box.

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
```diff
@@ -77,10 +78,11 @@ class TestingClass : Object() {
     var variantTest = Variant("test")
 
     @RegisterProperty
-    lateinit var resourceVisibleInEditor: Resource
+    lateinit var resourceVisibleInEditor: CustomResource //should be registered as Texture as Custom resource derives from texture
```
raniejade: I don't think this is correct example, adding a new resource type requires adding support to the engine. Plus we don't know how to instanstiate `CustomResource` at runtime if it has other parameters.

raniejade (changes requested):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
chippmann: That's not possible as the exact type is needed for the editor to show the correct inspector: https://github.com/godotengine/godot/blob/5f1107aa30295e686be6f41cb6d17fc2cff1e036/editor/editor_properties.cpp#L3287

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
chippmann: But then we have to actively prevent it. I'll change it this evening

chippmann: Did it now ^^

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
raniejade: I don't get it, we just use this to check if it's a resource right? Can we just check if the class inherits from `godot.Resource`?

chippmann: No.
This one checks if it's a resource at all:https://github.com/utopia-rise/godot-kotlin/blob/1a46bc80655f157f1fb998de0eb7452e18047eb7/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt#L12

This one checks if it's a registrable resource (if it does not derive from a godot resource, it IS a godot resource): https://github.com/utopia-rise/godot-kotlin/blob/1a46bc80655f157f1fb998de0eb7452e18047eb7/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt#L20

And this one gets the type of the resource we use for the typehint string: https://github.com/utopia-rise/godot-kotlin/blob/1a46bc80655f157f1fb998de0eb7452e18047eb7/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt#L24

And we need the godot resource type to be used in the hint so the editor knows which subinspector to show. If we don't provide it, it just shows the default subInspector for resources and not the individual once for textures, colors and so on.

raniejade: We can still provide that value without that list right? i.e just get a simple name from fqname of the class?

raniejade: On `fun KotlinType.isAnyResourceType()` the fact that it inherits from `Resource` then it should be registerable. Are we protecting against users declaring their own `Resource` types the engine does not support yet? My concern is that we have to maintain that list, when say there's a new resource type added by Godot.

chippmann: Yes with this impl we prevent that. And i think we should. Otherwise users have the problem @piiertho had this weekend with trying to register custom resources and nothing works the way it should.
So yes i get your point with the additional maintaining overhead, but we already have that in this file for all the core types.

piiertho: @chippmann the problem I had this weekend was a godot-cpp bug. Custom resources default values are not well initialized in godot-cpp. But it seems to work in D bindings. So it seems that GDNative allow user to define their own resource types.

chippmann: So we should allow it then and look at what the D-Binding has done?
And i think you are confusing something. Default values are set through code. What is set in the editor are not default values. And IIRC the editor does not support default values for resources at all.
What you probably mean is the cpp-binding has a bug with setting values that were defined in the editor (!= default values).

raniejade: I've been scouring the godot issues page and it looks like you can define custom resource? https://github.com/godotengine/godot/issues/27470 as an example. I guess they would inherit from `godot.Resource` directly as they would probably want to get reference counting support?

piiertho: In godot cpp you have to access it through `Ref<CustomResource>` in order to have reference counting and avoid dangling pointers.
@chippmann yes I'm talking about default values in code, that's buggy in cpp bindings.
@raniejade it does not show up in editor when trying to create a resource. The thing to do is to create a normal resource, and then manually point its script to gdns file. Then inspector will display properties you can set. This is due to the fact that a script does not really extends engine type.

raniejade: Hmmm, I think what we have now should suffice and lets see what users will say.

raniejade (approved):

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/extension/KotlinTypeExt.kt
chippmann: I agree
But back to the origins of this discussion; should we now prevent the registration of custom resources for now (which requires this list) or shouldn't we prevent it and possibly let user experience undefined behavior but less to maintain for us (remove this list)?

piiertho: @chippmann I think we should not prevent registration of custom resources, as other bindings does not do that

entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/provider/ResourceRegistrationValuesHandler.kt
```diff
@@ -15,19 +15,20 @@ class ResourceRegistrationValuesHandler(
         if (!propertyDescriptor.isLateInit && isVisibleInEditor()) {
             throw IllegalStateException("You initialized the property \"${propertyDescriptor.fqNameSafe}\". Properties of type Resource which are registered using the @RegisterProperty annotation and are visible in the editor are not allowed to have a default value. Use lateinit.")
         }
+        if (!propertyDescriptor.type.isAnyResourceType()) {
```
piiertho: Same comment, to me we should not prevent registration of custom resources.

piiertho (changes requested):
Don't know if we plan to enable custom resources here, otherwise I'm ok with everything !

piiertho (approved):

PR #143, opened 2020-04-28
Feature/vector

CedNaru:
Added Vector2
Added Vector3
Added Basis
Added Quat
Added dummy C structs inside godot.def to mimic the memory layout of the original cpp types. It makes setRawMemory/getRawMemory safe (replace the previous array access)
Added a typeAlias RealT for Double
Reimplemented isEqualApprox from Godot to handle float equality.

### DIFF
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
@@ -93,10 +93,6 @@ object Godot {
         godot_print_warning(description, function, file, line)
     }
 
-    internal fun printError(description: String, function: String, file: String, line: Int, exception: Exception) {
-        godot_print_error(description, function, file, line)
-    }
-
     internal fun printError(description: String, function: String, file: String, line: Int) {
         godot_print_error(description, function, file, line)
     }
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
@@ -84,4 +84,21 @@ object Godot {
         initHandle.compareAndSet(initHandle.value, 0)
         return ret
     }
+
+    internal fun print(message: String) {
+        godot_print(message.toGDString())
+    }
+
+    internal fun printWarning(description: String, function: String, file: String, line: Int) {
+        godot_print_warning(description, function, file, line)
+    }
+
+    internal fun printError(description: String, function: String, file: String, line: Int, exception: Exception) {
+        godot_print_error(description, function, file, line)
+        throw exception
+    }
+
+    internal fun printError(description: String, function: String, file: String, line: Int) {
+        godot_print_error(description, function, file, line)
+    }
 }
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
@@ -95,7 +95,6 @@ object Godot {
 
     internal fun printError(description: String, function: String, file: String, line: Int, exception: Exception) {
         godot_print_error(description, function, file, line)
-        throw exception
     }
 
     internal fun printError(description: String, function: String, file: String, line: Int) {
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
@@ -1,0 +1,30 @@
+package godot.core
+
+import kotlin.math.*
+
+const val CMP_EPSILON = 0.00001
+const val CMP_EPSILON2 = CMP_EPSILON * CMP_EPSILON
+const val Math_PI = 3.14159265358979323846
+
+const val PLANE_EQ_DOT_EPSILON = 0.999
+const val PLANE_EQ_D_EPSILON = 0.0001
+
+//taken from core/math/math_funcs.h in godot sources
+internal fun isEqualApprox(a: RealT, b: RealT, epsilon: Double = CMP_EPSILON): Boolean {
+    // Check for exact equality first, required to handle "infinity" values.
+    if (a == b) {
+        return true
+    }
+    // Then check for approximate equality.
+    var tolerance: RealT = epsilon * abs(a)
+    if (tolerance < epsilon) {
+        tolerance = epsilon
+    }
+    return abs(a - b) < tolerance
+}
+
+internal typealias RealT = Double
+
+fun Number.toRealT(): RealT {
+    return this.toDouble()
+}
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
@@ -4,13 +4,14 @@ import kotlin.math.*
 
 const val CMP_EPSILON = 0.00001
 const val CMP_EPSILON2 = CMP_EPSILON * CMP_EPSILON
-const val Math_PI = 3.14159265358979323846
 
 const val PLANE_EQ_DOT_EPSILON = 0.999
 const val PLANE_EQ_D_EPSILON = 0.0001
 
-//taken from core/math/math_funcs.h in godot sources
+
 internal fun isEqualApprox(a: RealT, b: RealT, epsilon: Double = CMP_EPSILON): Boolean {
+    //taken from https://github.com/godotengine/godot/blob/d8066aa6a4afb12ffddcec71bd7e051dcd04f3e1/core/math/math_funcs.h#L315
+
     // Check for exact equality first, required to handle "infinity" values.
     if (a == b) {
         return true
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
@@ -1,0 +1,617 @@
+@file:Suppress("unused", "MemberVisibilityCanBePrivate")
+
+package godot.core
+
+import godot.gdnative.godot_basis
+import godot.gdnative.godot_basis_layout
+import godot.gdnative.real_t
+import kotlinx.cinterop.*
+import kotlin.math.*
+
+
+class Basis(var x: Vector3, var y: Vector3, var z: Vector3) : CoreType {
+    //CONSTANTS
+    companion object {
+        val IDENTITY = Basis(1, 0, 0, 0, 1, 0, 0, 0, 1)
+        val FLIP_X = Basis(-1, 0, 0, 0, 1, 0, 0, 0, 1)
+        val FLIP_Y = Basis(1, 0, 0, 0, -1, 0, 0, 0, 1)
+        val FLIP_Z = Basis(1, 0, 0, 0, 1, 0, 0, 0, -1)
+
+        private val orthoBases: Array<Basis> =
+            arrayOf(
+                Basis(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0),
+                Basis(0.0, -1.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0),
+                Basis(-1.0, 0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0, 1.0),
+                Basis(0.0, 1.0, 0.0, -1.0, 0.0, 0.0, 0.0, 0.0, 1.0),
+                Basis(1.0, 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, 1.0, 0.0),
+                Basis(0.0, 0.0, 1.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0),
+                Basis(-1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 1.0, 0.0),
+                Basis(0.0, 0.0, -1.0, -1.0, 0.0, 0.0, 0.0, 1.0, 0.0),
+                Basis(1.0, 0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0, -1.0),
+                Basis(0.0, 1.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, -1.0),
+                Basis(-1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, -1.0),
+                Basis(0.0, -1.0, 0.0, -1.0, 0.0, 0.0, 0.0, 0.0, -1.0),
+                Basis(1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, -1.0, 0.0),
+                Basis(0.0, 0.0, -1.0, 1.0, 0.0, 0.0, 0.0, -1.0, 0.0),
+                Basis(-1.0, 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, -1.0, 0.0),
+                Basis(0.0, 0.0, 1.0, -1.0, 0.0, 0.0, 0.0, -1.0, 0.0),
+                Basis(0.0, 0.0, 1.0, 0.0, 1.0, 0.0, -1.0, 0.0, 0.0),
+                Basis(0.0, -1.0, 0.0, 0.0, 0.0, 1.0, -1.0, 0.0, 0.0),
+                Basis(0.0, 0.0, -1.0, 0.0, -1.0, 0.0, -1.0, 0.0, 0.0),
+                Basis(0.0, 1.0, 0.0, 0.0, 0.0, -1.0, -1.0, 0.0, 0.0),
+                Basis(0.0, 0.0, 1.0, 0.0, -1.0, 0.0, 1.0, 0.0, 0.0),
+                Basis(0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 1.0, 0.0, 0.0),
+                Basis(0.0, 0.0, -1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 0.0),
+                Basis(0.0, -1.0, 0.0, 0.0, 0.0, -1.0, 1.0, 0.0, 0.0)
+            )
+    }
+
+
+    //CONSTRUCTOR
+    constructor() :
+        this(Vector3(1, 0, 0), Vector3(1, 1, 0), Vector3(1, 0, 1))
+
+    constructor(
+        xx: Number,
+        xy: Number,
+        xz: Number,
+        yx: Number,
+        yy: Number,
+        yz: Number,
+        zx: Number,
+        zy: Number,
+        zz: Number
+    ) :
+        this(Vector3(xx, xy, xz), Vector3(yx, yy, yz), Vector3(zx, zy, zz))
+
+
+    constructor(from: Vector3) : this() {
+        setEuler(from)
+    }
+
+    constructor(quat: Quat) : this() {
+        val d = quat.lengthSquared()
+        val s = 2.0 / d
+        val xs = quat.x * s
+        val ys = quat.y * s
+        val zs = quat.z * s
+        val wx = quat.w * xs
+        val wy = quat.w * ys
+        val wz = quat.w * zs
+        val xx = quat.x * xs
+        val xy = quat.x * ys
+        val xz = quat.x * zs
+        val yy = quat.y * ys
+        val yz = quat.y * zs
+        val zz = quat.z * zs
+        set(
+            1.0 - (yy + zz), xy - wz, xz + wy,
+            xy + wz, 1.0 - (xx + zz), yz - wx,
+            xz - wy, yz + wx, 1.0 - (xx + yy)
+        )
+    }
+
+    constructor(axis: Vector3, phi: Double) : this() {
+        // Rotation matrix from axis and angle, see https://en.wikipedia.org/wiki/Rotation_matrix#Rotation_matrix_from_axis_and_angle
+        val axisq = Vector3(axis.x * axis.x, axis.y * axis.y, axis.z * axis.z)
+
+        val cosine: Double = cos(phi)
+        val sine: Double = sin(phi)
+
+        apply {
+            x.x = axisq.x + cosine * (1.0 - axisq.x)
+            x.y = axis.x * axis.y * (1.0 - cosine) - axis.z * sine
+            x.z = axis.z * axis.x * (1.0 - cosine) + axis.y * sine
+
+            y.x = axis.x * axis.y * (1.0 - cosine) + axis.z * sine
+            y.y = axisq.y + cosine * (1.0 - axisq.y)
+            y.z = axis.y * axis.z * (1.0 - cosine) - axis.x * sine
+
+            z.x = axis.z * axis.x * (1.0 - cosine) - axis.y * sine
+            z.y = axis.y * axis.z * (1.0 - cosine) + axis.x * sine
+            z.z = axisq.z + cosine * (1.0 - axisq.z)
+        }
+    }
+
+    internal constructor(native: CValue<godot_basis>) : this() {
+        memScoped {
+            this@Basis.setRawMemory(native.ptr)
+        }
+    }
+
+    internal constructor(mem: COpaquePointer) : this() {
+        this.setRawMemory(mem)
+    }
+
+
+    //INTEROP
+    override fun getRawMemory(memScope: MemScope): COpaquePointer {
+        val value = cValue<godot_basis_layout> {
+            x.x = this@Basis.x.x.toFloat()
+            x.y = this@Basis.x.y.toFloat()
+            x.z = this@Basis.x.z.toFloat()
+            y.x = this@Basis.y.x.toFloat()
+            y.y = this@Basis.y.y.toFloat()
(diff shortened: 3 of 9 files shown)


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -84,4 +84,21 @@ object Godot {
         initHandle.compareAndSet(initHandle.value, 0)
         return ret
     }
+
+    internal fun print(message: String) {
```
chippmann, line 88: What is the intention of defining this function internal?
Because IIRC we said we want the user to be able to use print functions as global functions and not from an object. We internally also use those functions; why not use the same ones? Is your idea that the global print function will just redirect the calls to these functions?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -84,4 +84,21 @@ object Godot {
...
+        godot_print(message.toGDString())
```
chippmann, line 89: Why is here a `toGDString` necessary and in the other places it's not?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -84,4 +84,21 @@ object Godot {
...
+    }
+
+    internal fun printWarning(description: String, function: String, file: String, line: Int) {
+        godot_print_warning(description, function, file, line)
+    }
+
+    internal fun printError(description: String, function: String, file: String, line: Int, exception: Exception) {
+        godot_print_error(description, function, file, line)
+        throw exception
```
raniejade: why do we need to throw?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
```diff
@@ -0,0 +1,20 @@
+package godot.core
+
+import kotlinx.cinterop.*
+
+
+internal interface CoreType {
+    fun getRawMemory(memScope: MemScope): COpaquePointer
+    fun setRawMemory(mem: COpaquePointer)
+    fun isNull(): Boolean = false
```
raniejade: when is this used?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
```diff
@@ -0,0 +1,30 @@
+package godot.core
+
+import kotlin.math.*
+
+const val CMP_EPSILON = 0.00001
+const val CMP_EPSILON2 = CMP_EPSILON * CMP_EPSILON
+const val Math_PI = 3.14159265358979323846
```
chippmann: Maybe add a comment on why you define PI yourself. Otherwise new contributors might ask themselve why we don't just use PI define in the math package of kotlin.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
```diff
@@ -0,0 +1,379 @@
...
+        Quat(x / f, y / f, z / f, w / f)
+
+    override fun equals(other: Any?): Boolean =
+        when (other) {
+            is Quat -> (x == other.x && y == other.y && z == other.z && w == other.w)
+            else -> false
+        }
+
+    override fun toString(): String {
+        return ""
```
raniejade: Implement it properly? :P

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
```diff
@@ -0,0 +1,30 @@
...
+
+const val PLANE_EQ_DOT_EPSILON = 0.999
+const val PLANE_EQ_D_EPSILON = 0.0001
+
+//taken from core/math/math_funcs.h in godot sources
```
chippmann: maybe add a perma link? as this file could been moved or modified

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -0,0 +1,616 @@
...
+
+    override fun setRawMemory(mem: COpaquePointer) {
+        val value = mem.reinterpret<godot_basis_layout>().pointed
+        x.setRawMemory(value.x.ptr)
+        y.setRawMemory(value.y.ptr)
+        z.setRawMemory(value.z.ptr)
+    }
+
+
+    //API
```
chippmann, line 155: What do you mean by `API`?
Edit: I suppose you used the comments from the godot source code; maybe make it a bit clearer or just remove the comments. IMHO the visibility should define the "api" functions

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -0,0 +1,616 @@
...
+
+    // getEulerXyz returns a vector containing the Euler angles in the format
+    // (a1,a2,a3), where a3 is the angle of the first rotation, and a1 is the last
+    // (following the convention they are commonly defined in the literature).
+    //
+    // The current implementation uses XYZ convention (Z is the first rotation),
+    // so euler.z is the angle of the (first) rotation around Z axis and so on,
+    //
+    // And thus, assuming the matrix is a rotation matrix, this function returns
+    // the angles in the decomposition R = X(a1).Y(a2).Z(a3) where Z(a) rotates
```
chippmann: Why not use KDoc?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -0,0 +1,616 @@
...
+            euler.x = atan2(this[0][1], this[1][1])
+            euler.y = Math_PI.toDouble() / 2.0
+            euler.z = 0.0
+        }
+        return euler
+    }
+
+    // getEulerYxz returns a vector containing the Euler angles in the YXZ convention,
+    // as in first-Z, then-X, last-Y. The angles for X, Y, and Z rotations are returned
+    // as the x, y, and z components of a Vector3 respectively.
```
chippmann: Use KDoc?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
```diff
@@ -0,0 +1,22 @@
...
+import godot.gdnative.*
+import kotlinx.cinterop.*
+
+
+fun String(mem: COpaquePointer) = mem.reinterpret<godot_string>().pointed.readValue().toKString()
+
+fun String.getRawMemory(memScope: MemScope): COpaquePointer =
+    cValue<godot_string> { godot_string_parse_utf8(this.ptr, this@getRawMemory) }.getPointer(memScope)
+
+internal fun String.toGDString(): CValue<godot_string> = cValue { godot_string_parse_utf8(this.ptr, this@toGDString) }
```
raniejade: Need to call `godot_string_destroy` when `godot_string` is not used anymore otherwise you will be leaking memory IIRC.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
```diff
@@ -0,0 +1,311 @@
...
+package godot.core
+
+import godot.gdnative.*
+import kotlinx.cinterop.*
+import kotlin.math.*
+
+class Vector2(var x: RealT, var y: RealT) : Comparable<Vector2>, CoreType {
+    //CONSTANTS
+    companion object {
+        const val AXIS_X: Int = 0
```
raniejade: Can we use enums for `AXIS_*` values?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -0,0 +1,616 @@
...
+    private fun getAxis(axis: Int): Vector3 =
+        Vector3(this[0][axis], this[1][axis], this[2][axis])
+
+    private fun setAxis(axis: Int, value: Vector3) {
+        this[0][axis] = value.x
+        this[1][axis] = value.y
+        this[2][axis] = value.z
+    }
+
+    fun rotated(axis: Vector3, phi: Double): Basis {
```
chippmann, line 433: Can't `PHI` be a const somewhere? IIRC `PHI` is always the same?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    fun setEuler(p_euler: Vector3) {
+        setEulerYxz(p_euler)
+    }
+
+    // setEulerXyz expects a vector containing the Euler angles in the format
+    // (ax,ay,az), where ax is the angle of rotation around x axis,
+    // and similar for other axes.
+    // The current implementation uses XYZ convention (Z is the first rotation).
```
chippmann: Make KDoc?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -0,0 +1,616 @@
...
+        val ret = xmat * (ymat * zmat)
+        this.x = ret.x
+        this.y = ret.y
+        this.z = ret.z
+    }
+
+    // setEulerYxz expects a vector containing the Euler angles in the format
+    // (ax,ay,az), where ax is the angle of rotation around x axis,
+    // and similar for other axes.
+    // The current implementation uses YXZ convention (Z is the first rotation).
```
chippmann: Kdoc?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
```diff
@@ -0,0 +1,311 @@
...
+        const val AXIS_Y: Int = 1
+        val ZERO = Vector2(0, 0)
```
raniejade: A problem here is that the user can accidentally mutate the instance of `ZERO` which is shared everywhere. A fix I had in my binding is to do:

```kotlin
val ZERO: Vector2
	get() = Vector2(0, 0)
```

This means that ZERO, will always return a unique instance.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector3.kt
```diff
@@ -0,0 +1,338 @@
...
+import godot.gdnative.godot_vector3
+import godot.gdnative.godot_vector3_layout
+import kotlinx.cinterop.*
+import kotlin.math.*
+
+
+class Vector3(var x: Double, var y: Double, var z: Double) : Comparable<Vector3>, CoreType {
+    //CONSTANTS
+    companion object {
+        const val AXIS_X: Int = 0
```
raniejade: enums for `AXIS_*`?

raniejade (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
```diff
@@ -0,0 +1,379 @@
...
+            temp[j] = (from[j][i] + from[i][j]) * s
+            temp[k] = (from[k][i] + from[i][k]) * s
+        }
+        set(temp[0], temp[1], temp[2], temp[3])
+    }
+
+    constructor(axis: Vector3, angle: RealT): this() {
+        val d: RealT = axis.length()
+        if (d == 0.0) set(0.0, 0.0, 0.0, 0.0)
+        else {
```
chippmann: I know you are not a fan of {} but i find this hard to read but do as you please

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    fun getEuler(): Vector3 {
+        return getEulerYxz()
+    }
+
+    // getEulerYxz returns a vector containing the Euler angles in the format
+    // (ax,ay,az), where ax is the angle of rotation around x axis,
+    // and similar for other axes.
+    // This implementation uses YXZ convention (Z is the first rotation).
```
chippmann: KDoc?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    fun setEuler(p_euler: Vector3) {
+        setEulerYxz(p_euler)
+    }
+
+    // setEulerXyz expects a vector containing the Euler angles in the format
+    // (ax,ay,az), where ax is the angle of rotation around x axis,
+    // and similar for other axes.
+    // This implementation uses XYZ convention (Z is the first rotation).
```
chippmann: KDoc?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
chippmann: Implementation missing (Intellij could auto generate this ;-)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
CedNaru: That's how it was before. I didn't think about changing that.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
```diff
@@ -0,0 +1,311 @@
...
+        return Vector2(round(x), round(y))
+    }
+
+    fun sign(): Vector2 {
+        return Vector2(sign(x), sign(y))
+    }
+
+    fun slerp(b: Vector2, t: RealT): Vector2 {
+        if (!this.isNormalized() || !b.isNormalized())
+            Godot.printError("Vectors not normalized", "slerp()", "Vector2.kt", 228)
```
chippmann: Again i know you are not a fan of {} but IIRC kotlin convetion is also {} for single lines :-)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
CedNaru: Not implemented yet, but initially generated classes inherit GodotObject that implement CoreType, in that case IsNull is overriden to check if the native object is created.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
CedNaru: Ha yeah I forgot that one.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
CedNaru: Yeah sorry, I just reused the previous one without changing anything. I needed it for some functions in the coretypes

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector3.kt
```diff
@@ -0,0 +1,338 @@
...
+        else 2
+    }
+
+    fun moveToward(to: Vector3, delta: RealT): Vector3 {
+        val vd = to - this
+        val len = vd.length()
+        return if (len <= delta || len < CMP_EPSILON)
+            to
+        else
+            this + vd / len * delta
```
chippmann: Everywhere {} but not here. I would keep it consistent :-)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
CedNaru: It was initially one but I tried to stick GDScript as much as possible. And there is no enum in GdScript as you already know.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector3.kt
```diff
@@ -0,0 +1,338 @@
...
+        return v2 * (v1.dot(v2) / v2.dot(v2))
+    }
+
+    fun reflect(by: Vector3): Vector3 {
+        return by - this * this.dot(by) * 2.0
+    }
+
+    fun rotated(axis: Vector3, phi: Double): Vector3 {
+        if (!axis.isNormalized())
+            Godot.printError("Axis not normalized", "rotated()", "Vector3.kt", 216)
```
chippmann: same as before

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector3.kt
```diff
@@ -0,0 +1,338 @@
...
+        return Vector3(round(x), round(y), round(z))
+    }
+
+    fun sign(): Vector3 {
+        return Vector3(sign(x), sign(y), sign(z))
+    }
+
+    fun slerp(b: Vector3, t: RealT): Vector3 {
+        if (!this.isNormalized() || !b.isNormalized())
+            Godot.printError("Vectors not normalized", "slerp()", "Vector3.kt", 239)
```
chippmann: again

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
CedNaru: Fair point.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
raniejade: maybe rename the method to `printAndThrowError`?

chippmann (changes requested):
Some very small things. All in all awesome work and thanks for the hard work!

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
raniejade: I don't think we need to limit ourselves with GDScript on this one.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
CedNaru: Exactly. Users shouldn't have direct access to printError for example.
The global print function are not the exact same ones.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
CedNaru: That's how it was before. I didn't touch that part. I guess I will convert them all

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
CedNaru: I guess you're right. I just paster the code from the C source. It just went over my head than PI was already in the Math package.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
CedNaru: It's just me who structured the code that way:
Constant
Constructors
API
Utilities

Not all public method are API functions. That's why I got a separate Utilities part for operator for example. 
I can remove them if you want. That's just how I do things.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
CedNaru: It's leftover from the previous bindings. I already removed a ton of them. I just forgot some of them. 
Don't worry I don't like it either.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
chippmann: :+1:

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
CedNaru: what do you mean by a perma link ?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
chippmann: No you don't have to, I'm fine with it if you prefer it that way.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
chippmann: :+1: Thanks a lot!

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
CedNaru: Conclusion: It just works !

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/GodotMath.kt
CedNaru: Done

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -95,7 +95,6 @@ object Godot {
 
     internal fun printError(description: String, function: String, file: String, line: Int, exception: Exception) {
```
chippmann: now the exception is unused. I would rename the method if we throw or print the stacktrace (maybe append it to the description?)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -159,15 +163,17 @@ class Basis(var x: Vector3, var y: Vector3, var z: Vector3) : CoreType {
...
+     * getEulerXyz returns a vector containing the Euler angles in the format
+     * (a1,a2,a3), where a3 is the angle of the first rotation, and a1 is the last
+     * (following the convention they are commonly defined in the literature).
+     *
+     * The current implementation uses XYZ convention (Z is the first rotation),
+     * so euler.z is the angle of the (first) rotation around Z axis and so on,
+     *
+     * And thus, assuming the matrix is a rotation matrix, this function returns
+     * the angles in the decomposition R = X(a1).Y(a2).Z(a3) where Z(a) rotates
+     */
```
chippmann, line 182: Maybe specify the return type with `@returns ...` same for all others i mentioned before

chippmann (changes requested):

chippmann (approved):
Thanks a lot for the docs as well! Great addition!

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import kotlinx.cinterop.*
+
+
+internal interface CoreType {
+    fun getRawMemory(memScope: MemScope): COpaquePointer
+    fun setRawMemory(mem: COpaquePointer)
+}
+
+fun Long.getRawMemory(memScope: MemScope): COpaquePointer {
```
piiertho: I think we need to merge our way to do, [this one](https://github.com/utopia-rise/godot-kotlin/blob/feature/api-code-generation-as-gradletask/godot-kotlin/godot-library/src/nativeInternal/kotlin/godot/internal/utils/getCValue.kt) has been added to api generation branch.
In any case, it should be internal.

piiertho (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
```diff
@@ -0,0 +1,434 @@
...
+    override fun toString(): String {
+        return "($x, $y)"
+    }
+
+    override fun hashCode(): Int {
+        return this.toString().hashCode()
+    }
+}
+
+operator fun RealT.times(vec: Vector2) = vec * this
```
raniejade, line 426: Why do we need this?

raniejade (commented):
One question, but overall LGTM!

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Vector2.kt
CedNaru: it's just the counterpart of     
```
operator fun times(rvalue: RealT): Vector2 =
        Vector2(x * rvalue, y * rvalue)
```
That way we can write: `2.0 * vec` or `vec * 2.0` regardless of the order

raniejade (approved):

piiertho (approved):

PR #144, opened 2020-05-04
Feature/GitHub ci

piiertho:
This adds build check to PR using github actions.

### DIFF
--- a/godot-kotlin/godot-library/build.gradle.kts
+++ b/godot-kotlin/godot-library/build.gradle.kts
@@ -1,7 +1,5 @@
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 
-val platform: String? by project
-
 plugins {
     kotlin("multiplatform")
     `maven-publish`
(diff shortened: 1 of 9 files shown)


godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -1,5 +1,7 @@
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 
+val platform: String? by project
```
raniejade: This will have an effect with publishing, as the `kotlinMultiplatform` publication will not contain information about the other platforms. I would suggest using the compile task for the specific platform, i.e `compileKotlinLinux` for linux instead of this platform parameter.

```
arguments: godot-library:${{ matrix.task }} -PignoreSamples
```

godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -9,9 +11,17 @@ plugins {
 kotlin {
     // we don't have godot-library in the mobile targets yet, limit these to desktop for now
     //has to be change in `GodotPlugin` of `godot-gradle-plugin` as well
-    macosX64("macos")
-    linuxX64("linux")
-    mingwX64("windows")
+    if (platform != null) {
```
raniejade: As mentioned above this is problematic. What your doing here is saying that `godot-library` only supports `platform`. During publication, regardless of what OS you are building for, the targets should be configured for _all_ platform that we support. This is required to publish multiplatform libraries.

raniejade (changes requested):

godot-kotlin/godot-library/build.gradle.kts
piiertho: right, that was causing trouble in former implem, I change it.

piiertho:
@raniejade and @chippmann ready to review !

raniejade (approved):
This is so good! LGTM!

PR #148, opened 2020-05-09
Feature/pool array (DONE)

CedNaru:
Added a draft of a PoolArray.

Just to discuss if this template is good enough to be implemented for every PoolXArray.

I used something close to RanieJade's implementation. The one we previously had seems to have memory leak and unnecessary copies.

### DIFF
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
@@ -8,6 +8,23 @@ internal interface CoreType {
     fun setRawMemory(mem: COpaquePointer)
 }
 
+abstract class NativeCoreType<C : CStructVar> : CoreType {
+    internal lateinit var _handle: CValue<C>
+}
+
+internal inline fun <T, reified C : CStructVar> callNative(
+    nativeCore: NativeCoreType<C>,
+    block: MemScope.(CPointer<C>) -> T
+): T {
+    return memScoped {
+        val ptr = nativeCore._handle.ptr
+        val ret: T = block(ptr)
+        nativeCore._handle = ptr.pointed.readValue()
+        ret
+    }
+}
+
+
 internal fun Long.getRawMemory(memScope: MemScope): COpaquePointer {
     return memScope.alloc<LongVar>().apply {
         this.value = this@getRawMemory
@@ -24,20 +41,4 @@ internal fun Boolean.getRawMemory(memScope: MemScope): COpaquePointer {
     return memScope.alloc<BooleanVar>().apply {
         this.value = this@getRawMemory
     }.ptr
-}
-
-abstract class NativeCoreType<C : CStructVar> : CoreType {
-    internal lateinit var _handle: CValue<C>
-}
-
-internal inline fun <T, reified C : CStructVar> callNative(
-    nativeCore: NativeCoreType<C>,
-    block: MemScope.(CPointer<C>) -> T
-): T {
-    return memScoped {
-        val ptr = nativeCore._handle.ptr
-        val ret: T = block(ptr)
-        nativeCore._handle = ptr.pointed.readValue()
-        ret
-    }
 }
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
@@ -8,6 +8,9 @@ internal interface CoreType {
     fun setRawMemory(mem: COpaquePointer)
 }
 
+/**
+ * WARNING: Do not inherit from this class, it is only there for the coretypes not entirely reimplemented in Kotlin.
+ */
 abstract class NativeCoreType<C : CStructVar> : CoreType {
     internal lateinit var _handle: CValue<C>
 }
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
@@ -25,3 +25,19 @@ internal fun Boolean.getRawMemory(memScope: MemScope): COpaquePointer {
         this.value = this@getRawMemory
     }.ptr
 }
+
+abstract class NativeCoreType<C : CStructVar> : CoreType {
+    internal lateinit var _handle: CValue<C>
+}
+
+internal inline fun <T, reified C : CStructVar> callNative(
+    nativeCore: NativeCoreType<C>,
+    block: MemScope.(CPointer<C>) -> T
+): T {
+    return memScoped {
+        val ptr = nativeCore._handle.ptr
+        val ret: T = block(ptr)
+        nativeCore._handle = ptr.pointed.readValue()
+        ret
+    }
+}
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
@@ -11,7 +11,7 @@ internal interface CoreType {
 /**
  * WARNING: Do not inherit from this class, it is only there for the coretypes not entirely reimplemented in Kotlin.
  */
-abstract class NativeCoreType<C : CStructVar> : CoreType {
+abstract class NativeCoreType<C : CStructVar> internal constructor() : CoreType {
     internal lateinit var _handle: CValue<C>
 }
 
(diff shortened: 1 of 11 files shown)


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolByteArray.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    override fun iterator(): Iterator<UByte> {
+        return IndexedIterator(size(), this::get)
+    }
+
+    internal inline fun <T> callNative(block: MemScope.(CPointer<godot_pool_byte_array>) -> T): T {
+        return memScoped {
+            val ptr = handle.ptr
+            val ret: T = block(ptr)
```
piiertho: Maybe checkNotNull can be called here to avoid duplicates

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolByteArray.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    override fun toString(): String {
+        return "PoolByteArray(${size()})"
+    }
+
+    override fun iterator(): Iterator<UByte> {
+        return IndexedIterator(size(), this::get)
+    }
+
+    internal inline fun <T> callNative(block: MemScope.(CPointer<godot_pool_byte_array>) -> T): T {
```
piiertho: If I remember well, we have other cores that use Native calls. Maybe this method should go in internal sourceset. But I understand that it is not the point of this PR as we are discussing the template.

piiertho (commented):
I really prefer this implem, having it as an Iterator will be far easier to use.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolByteArray.kt
CedNaru: Actually I made a mistake, It's not the result of the lambda that I should checkNotNull but the function we retrieve from Godot.gdnative.
Ex: 
 checkNotNull(Godot.gdnative.godot_X_fun)(args)
not  checkNotNull(Godot.gdnative.godot_X_fun(args))

It mean that the checkNotNull has to be defined inside the lambda.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolByteArray.kt
CedNaru: Hard to move this up. I need to access the internal handle. It's not part of the CoreType interface (and because it is internal I can't add it. Furthermore not all the CoreTypes have this property anyway)

chippmann:
I like the idea. I don't see a problem with it.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolByteArray.kt
piiertho: I don't get it. It is not possible to make this method static in internal sourceset ?

CedNaru: I can't use a static function for handling an instance property (_handle). It may be possible if the coretype itself is a parameter of the function but it doesn't work because _handle is not a property of the Coretype interface.
I tried to create a NativeCoreType abstract class that implement CoreType and add the _handle property but I still need to deal with the generic C structure. I tried something but couldn't get around this issue even with reified type parameters.

```
internal abstract class NativeCoreType<C : CStructVar> : CoreType {
    internal lateinit var _handle: CValue<C>
    
}

internal inline fun <T, reified C : CStructVar> callNative(
    nativeCore: NativeCoreType<C>,
    block: MemScope.(CPointer<C>) -> T
): T {
    return memScoped {
        val ptr = nativeCore._handle.ptr
        val ret: T = block(ptr)
        nativeCore._handle = ptr.pointed.readValue()
        ret
    }
}
```

image

piiertho: I tried to get it work by defining callNative in abstract class int 923c6d8.
But yes it is not static in this case.
Maybe I'm missing something about the problem ?
EDIT : Stupid thing I've done, we can define it in core type directly ...

CedNaru:
I tried to have NativeCall as a toplevel function. It works but I still had to implement a wrapper inside the poolArray class.
It works fine now, except the fact that I had to use an abstract class and it seems that a public class (PoolArray) can't inherit from an internal class (NativeCoreType). I had to make NativeCoreType a public abstract class :x. I am not fond of this solution.
image

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolByteArray.kt
```diff
@@ -155,12 +155,6 @@ class PoolByteArray(internal var handle: CValue<godot_pool_byte_array>) : CoreTy
     }
 
     internal inline fun <T> callNative(block: MemScope.(CPointer<godot_pool_byte_array>) -> T): T {
-        return memScoped {
-            val ptr = handle.ptr
-            val ret: T = block(ptr)
-            handle = ptr.pointed.readValue()
-            ret
-        }
+        return callNative(this, block)
```
piiertho, line 205: :+1:

piiertho:
> I tried to have NativeCall as a toplevel function. It works but I still had to implement a wrapper inside the poolArray class.
> It works fine now, except the fact that I had to use an abstract class and it seems that a public class (PoolArray) can't inherit from an internal class (NativeCoreType). I had to make NativeCoreType a public abstract class :x. I am not fond of this solution.
> image

Yep, I guess we cannot make better. I think the question is : "Is it a big problem that user can access this abstract class ?"

CedNaru:
Well it's an abstract class so it's not like an user could create an instance of it by mistake.
On the other side, a public abstract is like saying "You can implement your own coretype if you want" which can be misleading.

piiertho:
yep that's true, what @raniejade and @chippmann do you think guys ?

chippmann:
No i don't have a better idea
Let's slap a "dont inherit" kdoc on it and call it a day

CedNaru:
The remaining PoolArray are now done.
There are 4 methods in PoolByteArray that I can't implement as they are not available in the Gdnative API. 
For now they are there as private methods with a Todo().

CedNaru:
There were no equal methods either. I implemented a dirty one where I have to manually check each element individually. Probably slow but it works.

raniejade:
I think you can add the internal modifier to the primary constructor of `NativeCoreType`, that should prevent inheritance outside of the module.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolColorArray.kt
```diff
@@ -0,0 +1,187 @@
...
+     */
+    override fun equals(other: Any?): Boolean {
+        if (other is PoolColorArray) {
+            if (other.size() != this.size()) {
+                return false
+            }
+            val iter1 = this.iterator()
+            val iter2 = other.iterator()
+            while (iter1.hasNext()) {
+                if (iter1.next() != iter2.next())
```
raniejade: this will fail if `other` is shorter than `this`, can we convert the two iterators into a `List` first then do the comparisons using list equality. i.e `thisList == otherList`. I don't think it needs to be fast, as most likely the implementation will be a placeholder for the actual gdnative implementation.

raniejade (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/PoolColorArray.kt
CedNaru: Jus before I checked if they both have the same size so one can't be shorter than the other. 
But otherwise you are right. I guess I remade the wheel.

raniejade (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
```diff
@@ -8,6 +8,26 @@ internal interface CoreType {
     fun setRawMemory(mem: COpaquePointer)
 }
 
+/**
+ * WARNING: Do not inherit from this class, it is only there for the coretypes not entirely reimplemented in Kotlin.
+ */
+abstract class NativeCoreType<C : CStructVar> internal constructor() : CoreType {
```
chippmann, line 14: Just a suggestion: As this class is not intended to be inherited, i wonder if we should make that clear inside the className and not only in the kdoc. Something like `_NativeCoreType` or `InternalNativeCoreType`. Just a suggestion. I'm also fine with leaving it with the kdoc

chippmann (approved):

piiertho (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/CoreType.kt
CedNaru: I'll probably wait for all the CoreTypes to be implemented and then we will refactor details like this.

PR #149, opened 2020-05-09
Added Color

CedNaru:
Added Color CoreType with matching memory layout in Godot.def.


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,806 @@
...
+        }
+
+        this.r = r / 255.0
+        this.g = g / 255.0
+        this.b = b / 255.0
+        this.a = a / 255.0
+    }
+
+    constructor(from: Int) : this() {
+        TODO()
```
piiertho: Still TODO, is that wanted ?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,806 @@
...
+            res.b = (b * a * sa + over.b * over.a) / res.a
+        }
+        return res
+    }
+
+    /**
+     * Returns the most contrasting color.
+     */
+    fun contrasted(): Color {
+        val c = this
```
piiertho: Here I'm not sure if we should change this color or provide a copy.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,806 @@
...
+     * Returns the color’s grayscale representation.
+     * The gray value is calculated as (r + g + b) / 3.
+     */
+    fun gray(): RealT = (r + g + b) / 3.0
+
+    /**
+     * Returns the inverted color (1 - r, 1 - g, 1 - b, a).
+     */
+    fun inverted(): Color {
+        val c = this
```
piiertho: Same here.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    //Utilities
+    operator fun plus(c: Color) = Color(r + c.r, g + c.g, b + c.b, a + c.a)
+
+    operator fun minus(c: Color) = Color(r - c.r, g - c.g, b - c.b, a - c.a)
+
+    operator fun times(c: Color) = Color(r * c.r, g * c.g, b * c.b, a * c.a)
+
+    operator fun div(c: Color) = Color(r / c.r, g / c.g, b / c.b, a / c.a)
```
piiertho, line 781: Should we implement operators with real_t right operand, like here : https://github.com/godotengine/godot/blob/825a89d0dad7e69eabd5c431273e88a1ee74a75d/core/color.cpp#L585 ?

piiertho (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
CedNaru: I recreated the GDScript API:
[]( https://docs.godotengine.org/en/stable/classes/class_color.html?#class-color-method-color)

As all coreType are like primitive, all the methods return a modified copy of the original object. Unless you directly use the properties, you can't change the internal values of a non container coretype instance.

raniejade: This modifies the current object right?  Shouldn't it return a new copy?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
raniejade: this modifies the current object, I think we should return a new copy.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,811 @@
...
+            b + (1.0 - b) * amount,
+            a
+        )
+    }
+
+    /**
+     * Returns the linear interpolation with another color. The interpolation factor t is between 0 and 1.
+     */
+    fun linearInterpolate(otherColor: Color, t: RealT): Color {
+        val res = this
```
raniejade: same here, it modifies the current object not a copy. Note that `val res = this` won't create a new copy.

raniejade (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
CedNaru: Ho yeah it's true. My Bad

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,815 @@
...
+    }
+
+    //CONSTANTS
+    private enum class ColorByte(val size: Int, val shift: Int) {
+        BITS32(255, 8),
+        BITS64(65535, 16)
+    }
+
+    companion object {
+        val aliceblue: Color
```
raniejade: WDYT about making them inline?

```kotlin
        inline val aliceblue: Color
			get() = ...
```

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
```diff
@@ -0,0 +1,815 @@
...
+            b + (1.0 - b) * amount,
+            a
+        )
+    }
+
+    /**
+     * Returns the linear interpolation with another color. The interpolation factor t is between 0 and 1.
+     */
+    fun linearInterpolate(otherColor: Color, t: RealT): Color {
+        val res = Color(r, g,b, a)
```
raniejade: nit: formating: `Color(r, g, b, a)`

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
chippmann: i like the idea

CedNaru: done

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Color.kt
CedNaru: I always forget to format after small changes, sorry :p

raniejade (approved):

piiertho (approved):

chippmann (approved):

PR #150, opened 2020-05-10
Added RID and NodePath

CedNaru:
Added NodePath
Added RID

Those coretypes directly use the Godot Gdnative API. It would be hard to reimplement them in kotlin because they need access to some internal Godot variables. 
Ex: Godot use a global counter to assign an ID to an RID. Gdnative doesn't give access to this counter.

RID doesn't compile for now as it needs the Object Class for its constructor and I don't have access to this one because the coretype directory doesn't depend on Gen in the gradle build.
We can either change the gradle file or recreate the GodotObject interface (root of all the gen classes) we previously had. 
What do you think ?

### DIFF
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/NodePath.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/NodePath.kt
@@ -59,18 +59,28 @@ class NodePath : CoreType {
 
 
     //API
+    /**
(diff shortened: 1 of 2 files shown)


piiertho:
There is no dependency for now, but you should be able to access the object class from gen source dir, there's a dummy one to wait api generation is merged. We only have several source sets, so you can access any class you need in any sourceset.

EDIT : there's no dummy Object class, you can create one in gen source set to make it compile

chippmann (approved):

PR #151, opened 2020-06-01
Feature/bintray publish

piiertho:
Successfully deployed a [release](https://bintray.com/beta/#/utopia-rise/godot-kotlin-dev/godot-kotlin?tab=overview) on bintray with github actions.
I successfully resolved dependencies on mini game, but please try it on your side.
I think we still need to discuss about version convention, then everything would be done.
For now I set `0.1.0-3.2` as version, but what about alpha, rc, ... ?
Release happen on tag push.


godot-kotlin/godot-library/build.gradle.kts
```diff
@@ -46,4 +48,31 @@ tasks {
...
+                            override fun getDefaultExtension() = "module"
+                        })
+                    }
+                }
+            }
+        }
+    }
+}
+
+//TODO: See how to do with mobile platforms
```
raniejade, line 69: I think it is as simple as:

```
OS.LINUX -> arrayOf("kotlinMultiplatform", "metadata", "linux", "android64)
```

Meaning, on linux we also publish the `android64` target.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -1,5 +1,6 @@
 package godot.gradle
 
+import godot.compiler.plugin.CompilerPluginConst
```
raniejade: nit: unused import?

raniejade (approved):

godot-kotlin/godot-library/build.gradle.kts
piiertho: I would like to have those build in parrallel :)

raniejade: Ahh right, something that we can optimise later on. But as a start we can just build everything buildable for a particular os.

piiertho: Yep, I agree.
But as for beta we do not want arm platform, it is not imoportant yet

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
piiertho: oops, forgot that one

samples/mini-games/gradle.properties
```diff
@@ -2,4 +2,5 @@
 # kotlin.mpp.enableGranularSourceSetsMetadata=true
 # kotlin native compiler is now integrated with the gradle daemon, however, the default memory
 # settings for the daemon is not enough for kotlin native. This setting bumps the maximum memory to 3gb.
-org.gradle.jvmArgs=-Xmx3G
\ No newline at end of file
+org.gradle.jvmArgs=-Xmx3G
+godotVersion=3.2
```
chippmann: I would add a comment here stating in which file one needs to change the version as well because this is likely the place where one would change it and maybe forget or don't know about the other files where the version needs to be changed as well

godot-kotlin/godot-library/src/nativeGen/kotlin/godot/Resource.kt
```diff
@@ -1,5 +0,0 @@
-package godot
-
-//TODO: this needs to be properly implemented. It's just here for the implementation of the PropertyTypeHintAnnotations
-open class Resource {
```
chippmann: Because this was deleted, the sample cannot build atm. Either we restore this file or remove the corresponding testing lines in the test class inside the samples

godot-kotlin/godot-library/src/nativeGen/kotlin/godot/Texture.kt
```diff
@@ -1,5 +0,0 @@
-package godot
-
-//TODO: this needs to be properly implemented. It's just here for the implementation of the PropertyTypeHintAnnotations
-open class Texture : Resource() {
```
chippmann: Because this was deleted, the sample cannot build atm. Either we restore this file or remove the corresponding testing lines in the test class inside the samples

chippmann (changes requested):

godot-kotlin/godot-library/src/nativeGen/kotlin/godot/Resource.kt
piiertho: Oh yes, that's seems to be a git error I did when switching from api generation branch ^^'

samples/mini-games/gradle.properties
piiertho: That's not anymore needed, I forgot it.

chippmann (approved):

PR #152, opened 2020-06-30
Feature/transform

CedNaru:
Added Rect2
Added Transform
Added Transform2D
Updated godot.def with matching structs

### DIFF
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
@@ -2,10 +2,8 @@
 
 package godot.core
 
-import godot.gdnative.godot_aabb_layout
 import godot.gdnative.godot_rect2
 import godot.gdnative.godot_rect2_layout
-import godot.gdnative.real_t
 import kotlinx.cinterop.*
 import kotlin.math.max
 import kotlin.math.min
@@ -19,6 +17,21 @@ class Rect2(var position: Vector2, var size: Vector2) : CoreType {
             size = value - position
         }
 
+    //CONSTANTS
+    enum class Margin(val value: Int) {
+        LEFT(0),
+        TOP(1),
+        RIGHT(2),
+        BOTTOM(3)
+    }
+
+    companion object {
+        val MARGIN_LEFT = Margin.LEFT.value
+        val MARGIN_TOP = Margin.TOP.value
+        val MARGIN_RIGHT = Margin.RIGHT.value
+        val MARGIN_BOTTOM = Margin.BOTTOM.value
+    }
+
     //CONSTRUCTOR
     constructor() :
         this(Vector2(), Vector2())
@@ -26,7 +39,8 @@ class Rect2(var position: Vector2, var size: Vector2) : CoreType {
     constructor(x: RealT, y: RealT, width: RealT, height: RealT) :
         this(Vector2(x, y), Vector2(width, height))
 
-    internal constructor(native: CValue<godot_aabb_layout>) : this() {
+
+    internal constructor(native: CValue<godot_rect2>) : this() {
         memScoped {
             this@Rect2.setRawMemory(native.ptr)
         }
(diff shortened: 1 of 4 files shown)


piiertho (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
```diff
@@ -0,0 +1,263 @@
...
+            position.x - kotlin.math.min(size.x, 0.0),
+            kotlin.math.abs(size.x),
+            kotlin.math.abs(size.y)
+        )
+    }
+
+    /**
+     * Returns the intersection of this Rect2 and b.
+     */
+    fun clip(rect: Rect2): Rect2 {
```
chippmann: I would rename either `b` in the kdoc or the `rect` param so they match

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
```diff
@@ -0,0 +1,263 @@
...
+            point.y >= (position.y + size.y)    -> false
+            else                                -> true
+        }
+    }
+
+    /**
+     * Returns true if the Rect2 overlaps with b (i.e. they have at least one point in common).
+     * If include_borders is true, they will also be considered overlapping if their borders touch, even without intersection.
+     */
+    fun intersects(rect: Rect2): Boolean {
```
chippmann: Either the comment is not correct or a param is missing but there is no param boolean called `include_borders`

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
```diff
@@ -0,0 +1,263 @@
...
+     * Returns true if this Rect2 and rect are approximately equal, by calling is_equal_approx on each component.
+     */
+    fun isEqualApprox(rect: Rect2): Boolean {
+        return rect.position.isEqualApprox(this.position) && rect.size.isEqualApprox(this.size)
+    }
+
+    /**
+     * Returns a larger Rect2 that contains this Rect2 and b.
+     */
+    fun merge(rect: Rect2): Rect2 {
```
chippmann: Comment and param name do not match

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
```diff
@@ -0,0 +1,263 @@
...
+    //UTILITIES
+    override fun equals(other: Any?): Boolean {
+        return when (other) {
+            is Rect2 -> position == other.position && size == other.size
+            else -> false
+        }
+    }
+
+    override fun toString(): String {
+        return "$position, $size"
```
chippmann, line 265: Shouldn't we do something like `"Rect2[position = $position, size = $size]"` or something like that? otherwise it's hard to read the toString when put inside a log statement
Or is it like that in godot?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform.kt
```diff
@@ -0,0 +1,338 @@
...
+
+    override fun equals(other: Any?): Boolean {
+        return when (other) {
+            is Transform -> basis == other.basis && origin == other.origin
+            else -> false
+        }
+    }
+
+    override fun toString(): String {
+        return "$basis - $origin"
```
chippmann, line 330: Same question as with rect2

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform2D.kt
```diff
@@ -0,0 +1,358 @@
...
+        val x1 = tdoty(other.x)
+        val y0 = tdotx(other.y)
+        val y1 = tdoty(other.y)
+
+        return Transform2D(x0, x1, y0, y1, origin.x, origin.y)
+    }
+
+    override fun toString(): String {
+        return "${this.x}, ${this.y}, ${this.origin}"
+    }
```
chippmann, line 342: Same question as with Rect2

chippmann (changes requested):
Very cool thanks. Some minor questions though

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
CedNaru: image
This is like this in Godot. The toString function never return the type of the object.
We can discuss that this weekend to decide if we stick to Godot or replace it with a more specific string.
In the latter case, I will prepare a PR to fix all the core types (I will do one regardless because I noticed several mistakes I made in the previous cores).

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform.kt
CedNaru: same answer as with rect2

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform2D.kt
CedNaru: same answer as with rect2

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
CedNaru: Yeah sorry, I forgot to add it.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
chippmann: :+1:

chippmann (approved):
Thanks a lot for the work!

PR #154, opened 2020-07-04
Variant, Array and Dictionary

CedNaru:
Variant Class is implemented.
I tried to use the inline class feature of Kotlin. This would save us a lot of allocations as Variant is used to send/receive data to/from Godot.
Not many drawbacks to this method. The Kotlin Variant class only needs to keep a reference to the C++ pointer. But for now, inline classes don't allow secondary constructor so I had to create fake ones as normal functions.
I'm not sure if it will work. The compiler is complaining about constructor ambiguity for now.

I am also not sure about all the wrap/unwrap functions. Unwrap, as it is now, seems really unsafe. I tried to change the signature of the function but if I do that, I break another part of the library outside the cores. So I'm open to suggestions.


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Variant.kt
```diff
@@ -0,0 +1,585 @@
...
+fun Variant(from: PoolColorArray) = Variant.buildAny(Godot.gdnative.godot_variant_new_pool_color_array, from)
+fun Variant(from: PoolIntArray) = Variant.buildAny(Godot.gdnative.godot_variant_new_pool_int_array, from)
+fun Variant(from: PoolRealArray) = Variant.buildAny(Godot.gdnative.godot_variant_new_pool_real_array, from)
+fun Variant(from: PoolStringArray) = Variant.buildAny(Godot.gdnative.godot_variant_new_pool_string_array, from)
+fun Variant(from: PoolVector2Array) = Variant.buildAny(Godot.gdnative.godot_variant_new_pool_vector2_array, from)
+fun Variant(from: PoolVector3Array) = Variant.buildAny(Godot.gdnative.godot_variant_new_pool_vector3_array, from)
+fun <T> Variant(from: VariantArray<T>) = Variant.buildAny(Godot.gdnative.godot_variant_new_array, from)
+
+//Generic constructor to wrap objects unrelated to Godot into a null variant
+fun Variant(from : Any?) = Variant()
```
raniejade: I think this might cause some ambiguity errors, passing an `AABB` for example can match `fun Variant(from: AABB)` and `fun Variant(from: Any?)`

CedNaru: Yeah, it does.
But I had to put this constructor here because there is a dependency chain in the library. This constructor was is needed because the VariantArray placeholder is using it with 
`    fun T.toVariant() = Variant(this)`.
It's necessary to cast the generic type of the VariantArray into a Variant

```
    fun add(index: Int, element: T) {
        callNative {
            checkNotNull(Godot.gdnative.godot_array_insert)(it, index, element.toVariant()._handle.ptr )
        }
    }
```

That's one of the problems I got dealing with a generic VariantArray.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Dictionary.kt
```diff
@@ -43,11 +49,277 @@ class Dictionary<T> : NativeCoreType<godot_dictionary>, Iterable<T> {
...
+    override fun toString(): String {
+        return "Dictionary($size)"
+    }
+
     internal inline fun <T> callNative(block: MemScope.(CPointer<godot_dictionary>) -> T): T {
         return callNative(this, block)
     }
+
+    override fun iterator(): Iterator<Pair<Variant, Variant>> {
+        TODO("Not yet implemented")
```
chippmann: Missing impl?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform2D.kt
```diff
@@ -192,7 +192,9 @@ class Transform2D(var x: Vector2, var y: Vector2, var origin: Vector2) : CoreTyp
      * Returns true if this transform and transform are approximately equal, by calling is_equal_approx on each component.
      */
     fun isEqualApprox(transform: Transform2D): Boolean {
-        return transform.x.isEqualApprox(this.x) && transform.y.isEqualApprox(this.y) && transform.origin.isEqualApprox(this.origin)
+        return transform.x.isEqualApprox(this.x) && transform.y.isEqualApprox(this.y) && transform.origin.isEqualApprox(
+            this.origin
+        )
```
chippmann: Really minor detail but since everything else is pretty perfectly formatted, i would format this as well so the `&&` are on a new line or something similar

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/CoreVariantArray.kt
```diff
@@ -0,0 +1,419 @@
...
+    constructor() : super()
+    constructor(other: AABBArray) : super(other)
+    internal constructor(native: CValue<godot_array>) : super(native)
+    internal constructor(mem: COpaquePointer) : super(mem)
+
+    override fun getCore(value: Variant): AABB = value.asAABB()
+    override fun getCoreArray(value: CValue<godot_array>) = AABBArray(value)
+}
+
+fun AABBArrayOf(vararg elements: AABB) = AABBArray().also {
```
chippmann: I know i'll introduce a ton of work with my proposal here, but maybe it's a good idea:
How about an additional extension function like this:
`fun Iterable<AABB>.toVariantArray(): AABBArray`

chippmann: Of course we would need that for all array types...

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/GodotArray.kt
```diff
@@ -0,0 +1,277 @@
+package godot.core
+
+import godot.gdnative.godot_array
+import kotlinx.cinterop.*
+import platform.posix.VISIT
+
+abstract class GodotArray<T> : NativeCoreType<godot_array>(), Iterable<T> {
```
chippmann: Maybe add a kDoc that this is also a class from which one should not derive?

chippmann: Or maybe one CAN derive from it to create custom godot array classes like one can in gdscript?
Not sure... what's your oppinion?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/pool/PoolByteArray.kt
```diff
@@ -147,28 +147,28 @@ class PoolByteArray : NativeCoreType<godot_pool_byte_array>, Iterable<UByte> {
      * Not available in the Gdnative API and no workaround for now
      */
     private fun compress(compressionMode: Int = 0): PoolByteArray {
-        TODO()
+        return PoolByteArray()
     }
```
chippmann: I would not just return a empty array. I would let it crash and state so in the kDoc with `@throws` otherwise this method just "works" and the user might search for the bug for a long time until he realizes that this method in reality does not "work"

chippmann (changes requested):
Very awesome work! Thanks a lot. Some minor suggestions though.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/pool/PoolByteArray.kt
CedNaru: Right, I'll do that.
I also considered removing them but I want them around to notify users that we didn't just forget those methods, just that we can't implement them.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/GodotArray.kt
CedNaru: It's fine to derive them. It won't break anything. Actually on another project using C++ bindings, I had to inherit from Array to do my own container so it helped me a lot. (I cheated by storing a C pointer as a Int inside a Godot Array because I didn't want to use the STL)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/CoreVariantArray.kt
CedNaru: I'm not sure about that because that would probably something we'll have to remove for the next 4.0 version.
Today(3.2) it's fine because behind the hood everything is a cpp VariantArray, it's just wrapped by a generic kotlin object. We just need to take the CValue and put it into a  different kotlin typed Array.
But soon, those two will be different objects and we'll need a wrap the elements into a Variant one by one and insert them into the VariantArray which could be slow unless GDNative provide us the method directly.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform2D.kt
CedNaru: It's supposed to be formatted as well. Maybe the linter only use new lines where there is more than 3 arguments. I will fix that manually.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Dictionary.kt
CedNaru: Yeah sorry. I forgot it

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform2D.kt
chippmann: Yeah the linter only breaks on certain lenghts and then only to the next "breakable" point. Lines like that mostly need manual formatting

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/CoreVariantArray.kt
chippmann: Yes that's true but either we provide this extension function for the user or he does it by himselve when he uses kotlin iterables internally and passes them to godot. In any way there is a performance penalty. If we do it for the user or if he does it by himselve

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/GodotArray.kt
chippmann: :+1:

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/CoreVariantArray.kt
piiertho: I agree with @chippmann .
We will have to modify a lot bindings for 4.0. But for now we're on 3.2 so we need to provide user friendly bindings for 3.2

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/pool/PoolByteArray.kt
piiertho: What about `GDNativeNotDefinedException` ? :)

CedNaru: [It exists already, but I can"t inherit from it. It's not an open class](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-not-implemented-error/)

piiertho (commented):
LGTM, once other spotted problems are fixed.
Only have some fear about varargs call
```kotlin
internal fun _icall_varargs(
  mb: CPointer<godot_method_bind>,
  inst: COpaquePointer,
  arguments: Array<*>
): Variant {
  memScoped {
          val args = allocArray<CPointerVar<godot_variant>>(arguments.size)
          for ((i,arg) in arguments.withIndex()) args[i] = Variant.wrap(arg).handle.ptr
          val result = Godot.gdnative.godot_method_bind_call!!.invoke(mb, inst, args,
          arguments.size, null)
          for (i in arguments.indices) Godot.gdnative.godot_variant_destroy!!.invoke(args[i])
          return Variant(result)
      }

}
```

called here:
image
As we don't know which types are variadic arguments we'll have to unwrap them. Don't know how it will perform.

piiertho (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/core/AABBArray.kt
```diff
@@ -0,0 +1,35 @@
...
+
+@ExperimentalUnsignedTypes
+fun AABBArrayOf(vararg elements: AABB) = AABBArray().also {
+    for (arg in elements) {
+        it.append(arg)
+    }
+}
+
+@ExperimentalUnsignedTypes
+fun Iterable<AABB>.toVariantArray() = AABBArray().also{
```
chippmann: As we discussed. This function will be a performance problem. I suggest we add a kDoc to all the `.toVariantArray()` functions stating this does not perform well and if the user wants good performance he has to use variantArray directly.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/core/AABBArray.kt
```diff
@@ -0,0 +1,35 @@
...
+    for (arg in this) {
+        it.append(arg)
+    }
+}
+
+@ExperimentalUnsignedTypes
+fun AABBArray(iter: Iterable<AABB>) = iter.toVariantArray()
```
chippmann, line 45: Same with this function. I really like this idea of you. But i would ad the kDoc here as well as this "constructor" will not perform well, but it's needed.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/core/AABBArray.kt
```diff
@@ -0,0 +1,35 @@
...
+    internal constructor(native: CValue<godot_array>) : super(native)
+    internal constructor(mem: COpaquePointer) : super(mem)
+
+    override fun getCore(value: Variant): AABB = value.asAABB()
+    override fun getCoreArray(value: CValue<godot_array>) =
+        AABBArray(value)
+}
+
+@ExperimentalUnsignedTypes
+fun AABBArrayOf(vararg elements: AABB) = AABBArray().also {
```
chippmann, line 23: Same here as well

chippmann (changes requested):

chippmann (approved):

piiertho (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/IndexedIterator.kt
```diff
@@ -13,3 +13,20 @@ internal class IndexedIterator<T>(
         return getter(index++)
     }
 }
+
+data class Entry<out K, out V>(val key: K, val value: V)
```
raniejade, line 17: why not use `Pair`?

CedNaru: I first used Pair. But the properties to access the two elements are .first and .second.
Entry does use .key and .value

PR #159, opened 2020-07-15
Global functions

CedNaru:
So far I implemented all the maths functions. Some are just wrappers for kotlin.math, some have to be reimplemented.
Some functions won't be implemented because they rely heavily on the dynamic nature of GdScript and it would be a pain to replicate that in Kotlin.
For now, I am waiting for the gen class in order to implement the random and resource loading functions. 

There is also a bunch of print functions inside Gdscript, I don't know if it's worth to implement them.

[Reference](https://docs.godotengine.org/en/3.2/classes/class_@gdscript.html)

### DIFF
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
@@ -4,10 +4,7 @@ package godot.core
 
 import godot.gdnative.godot_quat
 import godot.gdnative.godot_quat_layout
-import godot.internal.type.CMP_EPSILON
-import godot.internal.type.CoreType
-import godot.internal.type.RealT
-import godot.internal.type.toRealT
+import godot.internal.type.*
 import kotlinx.cinterop.*
 import kotlin.math.*
 
@@ -175,10 +172,10 @@ class Quat(var x: RealT, var y: RealT, var z: RealT, var w: RealT) : CoreType {
      * Returns true if this quaterion and quat are approximately equal, by running isEqualApprox on each component.
      */
     fun isEqualApprox(other: Quat): Boolean {
-        return godot.internal.type.isEqualApprox(other.x, x)
-        godot.internal.type.isEqualApprox(other.y, y)
-        godot.internal.type.isEqualApprox(other.z, z)
-        godot.internal.type.isEqualApprox(other.w, w)
+        return isEqualApprox(other.x, x)
+        isEqualApprox(other.y, y)
+        isEqualApprox(other.z, z)
+        isEqualApprox(other.w, w)
     }
 
     /**
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
@@ -7,6 +7,7 @@ import godot.gdnative.godot_quat
 import godot.gdnative.godot_quat_layout
 import kotlinx.cinterop.*
 import kotlin.math.*
+import godot.internal.*
 
 class Quat(var x: RealT, var y: RealT, var z: RealT, var w: RealT) : CoreType {
     //CONSTANTS
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
@@ -2,12 +2,14 @@
 
 package godot.core
 
-import godot.internal.type.CoreType
 import godot.gdnative.godot_quat
 import godot.gdnative.godot_quat_layout
+import godot.internal.type.CMP_EPSILON
+import godot.internal.type.CoreType
+import godot.internal.type.RealT
+import godot.internal.type.toRealT
 import kotlinx.cinterop.*
 import kotlin.math.*
-import godot.internal.*
 
 class Quat(var x: RealT, var y: RealT, var z: RealT, var w: RealT) : CoreType {
     //CONSTANTS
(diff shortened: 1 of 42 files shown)


godot-kotlin/godot-library/src/nativePublic/kotlin/godot/global/gd.kt
```diff
@@ -0,0 +1,62 @@
...
+            Variant.Type.DICTIONARY -> len(what.asDictionary())
+            Variant.Type.ARRAY -> len(what.asVariantArray())
+            Variant.Type.POOL_BYTE_ARRAY -> len(what.asPoolByteArray())
+            Variant.Type.POOL_INT_ARRAY -> len(what.asPoolIntArray())
+            Variant.Type.POOL_REAL_ARRAY -> len(what.asPoolRealArray())
+            Variant.Type.POOL_STRING_ARRAY -> len(what.asPoolStringArray())
+            Variant.Type.POOL_VECTOR2_ARRAY -> len(what.asPoolVector2Array())
+            Variant.Type.POOL_VECTOR3_ARRAY -> len(what.asPoolVector3Array())
+            Variant.Type.POOL_COLOR_ARRAY -> len(what.asPoolColorArray())
+            else -> throw IllegalArgumentException("Variant can not provide a length.")
```
chippmann: Shouldn't we in this case return -1?

CedNaru: The Gdscript doc says that it throws a fatal error when the variant type is not right. Just trying to be consistent with the original API.

chippmann: Ah ok. Then it's fine for me.

godot-kotlin/godot-library/src/nativePublic/kotlin/godot/global/gd.kt
```diff
@@ -0,0 +1,84 @@
...
+import godot.Object
+import godot.RandomNumberGenerator
+import godot.global.gdPrint
+import godot.global.gdResource
+import kotlinx.cinterop.invoke
+import kotlinx.cinterop.memScoped
+import kotlin.test.assertTrue
+
+
+object gd : gdMath, gdCore, gdRandom, gdPrint, gdResource {
```
chippmann: Why is `gd` lowercase?

chippmann (approved):
LGTM. Thanks very much for the work! Especially the kDocs!
One question i have but nothing that prevents this from merging

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
```diff
@@ -173,9 +173,9 @@ class Quat(var x: RealT, var y: RealT, var z: RealT, var w: RealT) : CoreType {
      */
     fun isEqualApprox(other: Quat): Boolean {
         return isEqualApprox(other.x, x)
-            && isEqualApprox(other.y, y)
-            && isEqualApprox(other.z, z)
-            && isEqualApprox(other.w, w)
+        isEqualApprox(other.y, y)
```
piiertho: I think it is an attention error. It is unreachable.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
```diff
@@ -4,7 +4,7 @@ package godot.core
 
 import godot.gdnative.godot_rect2
 import godot.gdnative.godot_rect2_layout
-import godot.internal.type.CoreType
+import godot.internal.type.*
```
piiertho, line 7: Do we need the star import ? Or do we only need CoreType ?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform.kt
```diff
@@ -4,7 +4,7 @@ package godot.core
 
 import godot.gdnative.godot_transform
 import godot.gdnative.godot_transform_layout
-import godot.internal.type.CoreType
+import godot.internal.type.*
```
piiertho, line 7: Same

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform2D.kt
```diff
@@ -4,14 +4,15 @@ package godot.core
 
 import godot.gdnative.godot_transform2d
 import godot.gdnative.godot_transform2d_layout
-import godot.internal.type.CoreType
+import godot.internal.type.*
```
piiertho, line 7: same

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Variant.kt
```diff
@@ -5,13 +5,12 @@ import godot.Object
 import godot.gdnative.godot_array
 import godot.gdnative.godot_variant
 import godot.gdnative.godot_variant_type
-import godot.internal.type.CoreType
+import godot.internal.type.*
```
piiertho, line 7: Same.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/VariantArray.kt
```diff
@@ -2,7 +2,7 @@ package godot.core
 
 import godot.Object
 import godot.gdnative.godot_array
-import godot.internal.type.CoreType
+import godot.internal.type.*
```
piiertho, line 5: Same.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/array/core/CoreArray.kt
```diff
@@ -2,7 +2,7 @@ package godot.core
 
 import godot.Object
 import godot.gdnative.godot_array
-import godot.internal.type.CoreType
+import godot.internal.type.*
```
piiertho, line 5: Same.

piiertho (changes requested):
Few changes and comments, nothing to say on everything else :)
Thanks !

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Rect2.kt
CedNaru: I changed this package. Now it contains the NaturalT/RealT alias and the isEqualApprox() function. Not putting the star and just optimizing the imports, add a ton of individual imports and  modify stuff like 
`godot.internal.type.isEqualApprox()`

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Quat.kt
CedNaru: Ha yeah. Like I say below. The optimized imported messed up that part of the code, I probably made a mistake when rewriting it.

piiertho (approved):
Thanks !

PR #163, opened 2020-07-17
add an EnumArray

CedNaru:
Added an EnumArray to the coreType.
Behind the hood, it's just a normal IntVariantArray but it requires the user to provide a mapping function from Int to the generic Enum type.


chippmann (approved):
Tanks a lot for this impl!
Looks good to me

PR #164, opened 2020-07-21
[Proposal] godot-extension-injection

chippmann:
When we started the merge we talked about possible extensions that a user can use together with this binding.

This is a proposal for such an extension. It is chosen over the coroutines as it covers much more than just code extension for the binding (which would be quite simple to implement).
It would cover:
- Code extensions for the user to use (Annotations, new code)
- Entry generation
- Annotation Processing

## Goal:
The user is able to directly inject nodes into properties without the need of `getNode(NodePath("some/node/Path"))`

This is achieved by introducing a new annotation: `@Inject`

- With it a user marks a property he wants to have injected.
- And calls the function `inject` inside the `_ready` function on a `Injector` class we generate
- We then generate a corresponding property of type `NodePath` and expose it to the editor.
- Inside the editor the user can drac and drop a node to the `NodePath` field in the inspector.
- When the game runs, Godot calls the setter of the `NodePath` property we exposed to the editor.
- With that `NodePath` we call `getNode` internally when the ready function is called

Upsides:
The user does not have to write all the time:
```kotlin
@RegisterProperty
lateinit var nodePath: NodePath
lateinit var node: Node
@RegisterProperty
lateinit var nodePath2: NodePath
lateinit var node2: Node
@RegisterProperty
lateinit var nodePath3: NodePath
lateinit var node3: Node

override fun _ready() {
    super._ready()
    node = getNode(nodePath)
    node2 = getNode(nodePath2)
    node3 = getNode(nodePath3)
}
```

he can just write:
```kotlin
@Inject
lateinit var node: Node
@Inject
lateinit var node2: Node
@Inject
lateinit var node3: Node

override fun _ready() {
    super._ready()
    JumpTableInjector.injectNodes(this) //injection happens here
}
```

For classes with many nodes that are injected with this approach. The first common way can introduce a lot of boilerplate code really fast

## How it should work
What will be generated is in the `TestingClass` for my convenience documented with comments in code

What is not yet present in this draft PR; the entry generation part
We would need to discuss how we want to implement the support for extensions for the `entry-generator` and the `annotation-processor`.
One way could be to load extension jars during entry generation runtime using reflection, but that's just an idea

### DIFF
--- a/godot-kotlin/godot-extension-injection/src/native/kotlin/godot/extension/injection/InjectionHelperPropertyHandler.kt
+++ b/godot-kotlin/godot-extension-injection/src/native/kotlin/godot/extension/injection/InjectionHelperPropertyHandler.kt
@@ -1,0 +1,17 @@
+package godot.extension.injection
+
+import godot.core.NodePath
+import godot.core.Variant
+import kotlin.reflect.KMutableProperty0
+
+class InjectionHelperPropertyHandler(private val nodePath: KMutableProperty0<NodePath>) {
+    fun get(): Variant {
+        return Variant.wrap(
+            nodePath.get()
+        )
+    }
+
+    fun set(value: Variant) {
+        nodePath.set(value.unwrap())
+    }
+}
--- a/godot-kotlin/godot-extension-injection/build.gradle.kts
+++ b/godot-kotlin/godot-extension-injection/build.gradle.kts
@@ -1,0 +1,65 @@
+import com.jfrog.bintray.gradle.tasks.BintrayUploadTask
+import org.gradle.api.publish.maven.internal.artifact.FileBasedMavenArtifact
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+
+plugins {
+    kotlin("multiplatform")
+    `maven-publish`
+}
+
+kotlin {
+    // we don't have godot-library in the mobile targets yet, limit these to desktop for now
+    //has to be change in `GodotPlugin` of `godot-gradle-plugin` as well
+    macosX64("macos")
+    linuxX64("linux")
+    mingwX64("windows")
+
+    targets.withType<KotlinNativeTarget> {
+        compilations.getByName("main") {
+            defaultSourceSet {
+                kotlin.srcDir("src/native/kotlin")
+            }
+            dependencies {
+                implementation("com.utopia-rise:godot-library:${version}")
+            }
+        }
+    }
+    sourceSets {
+        all {
+            languageSettings.enableLanguageFeature("InlineClasses")
+        }
+    }
+}
+
+tasks {
+    build {
+        finalizedBy(publishToMavenLocal)
+    }
+
+    // workaround to upload gradle metadata file
+    // https://github.com/bintray/gradle-bintray-plugin/issues/229
+    withType<BintrayUploadTask> {
+        doFirst {
+            publishing.publications.withType<MavenPublication> {
+                buildDir.resolve("publications/$name/module.json").also {
+                    if (it.exists()) {
+                        artifact(object: FileBasedMavenArtifact(it) {
+                            override fun getDefaultExtension() = "module"
+                        })
+                    }
+                }
+            }
+        }
+    }
+}
+
+//TODO: See how to do with mobile platforms
+project.extra["artifacts"] = when (currentOS) {
+    OS.LINUX -> arrayOf("kotlinMultiplatform", "metadata", "linux")
+    OS.WINDOWS -> arrayOf("windows")
+    OS.MACOS -> arrayOf("macos")
+}
+
+apply {
+    plugin(BintrayPublish::class.java)
+}
--- a/godot-kotlin/godot-extension-injection/src/native/kotlin/godot/extension/injection/ClassBuilderExt.kt
+++ b/godot-kotlin/godot-extension-injection/src/native/kotlin/godot/extension/injection/ClassBuilderExt.kt
@@ -1,0 +1,28 @@
+package godot.extension.injection
+
+import godot.Object
+import godot.core.ClassBuilder
+import godot.core.NodePath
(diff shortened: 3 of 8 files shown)


chippmann:
This has by no means any priority. It's just here to be discussed as we need to discuss it anyways if we really want to have extensions.

chippmann:
Relevant part from the `TestingClass` inline documentation (as the diff is quite unreadable):
```kotlin
@RegisterClass
class CustomClass : Node() {

    @Inject("SomeCustomNodePathName", "Some/Path/ToNode")
    lateinit var nodeToInject: Tween
    @Inject
    lateinit var otherNodeToInject: Tween

    override fun _ready() {
        super._ready()
        JumpTableInjector.injectNodes(this) //injection happens here
    }

    //...
}

//==================================================================
//Start EntryGenerationFile

//with godot-extension-injection we generate

//generate once
/*
Uses a jump table
+ convenience as user can always use the same line for injection
- runtime perf lower because of type lookup
- Only available after first build (but compiles also for the first build if typed correctly)
 */
object JumpTableInjector {
    fun <T: Node> injectNodes(instance: T) {
        when (instance) {
            is CustomClass -> CustomClassInjector.injectNodes(instance)
        }
    }
}

//for each class that has at least one @Inject annotation
/*
Used by jump table
User can also directly use this for less convenience but slightly better performance
+ Perf
- Cumbersome, as to use it for each new class the user has to make a build first for the ide to recognize it (but compiles also for the first time if typed correctly)
 */
object CustomClassInjector : NodeInjector<CustomClass> {
    var nodeToInjectNodePath = NodePath()

    override fun injectNodes(instance: CustomClass) {
        instance.nodeToInject = instance.getNode(nodeToInjectNodePath) as Tween
    }
}


@CName("godot_nativescript_init")
fun NativeScriptInit(handle: COpaquePointer) {
    Godot.nativescriptInit(handle)
    with(ClassRegistry(handle)) {
        registerClass("example.TestingClass", "Node", ::CustomClass, false) {

            //generated with godot-extension-injection
            injectionHelper("SomeCustomNodePathName", CustomClassInjector::nodeToInjectNodePath, "Some/Path/ToNode") //note the custom entries for the NodePath name and default value

            //generated using normal entry generation
            property(
                CustomClass::nodeToInject.name,
                CustomClass::nodeToInject,
                Variant.Type.OBJECT,
                rpcMode = RPCMode.DISABLED
            )

            //generated with godot-extension-injection
            injectionHelper("otherNodeToInjectNodePath", CustomClassInjector::nodeToInjectNodePath) //nodePath name is (property name + NodePath) suffix

            //generated using normal entry generation
            property(
                CustomClass::otherNodeToInject.name,
                CustomClass::otherNodeToInject,
                Variant.Type.OBJECT,
                rpcMode = RPCMode.DISABLED
            )
        }
    }
}

//End EntryGenerationFile
//==================================================================
```

godot-kotlin/godot-extension-injection/src/native/kotlin/godot/extension/injection/InjectionHelperPropertyHandler.kt
```diff
@@ -0,0 +1,17 @@
+package godot.extension.injection
+
+import godot.core.NodePath
+import godot.core.Variant
+import kotlin.reflect.KMutableProperty0
+
+class InjectionHelperPropertyHandler(private val nodePath: KMutableProperty0<NodePath>) {
+    fun get(): Variant {
+        return Variant.wrap(
+            nodePath.get()
```
CedNaru, line 10: In this kind of case where you know the type of the instance (NodePath in this case) you should just write :
`nodePath.get().toVariant()`
Wrap() is for a generic T Object so it's slower because we have to check the type before.

godot-kotlin/godot-extension-injection/src/native/kotlin/godot/extension/injection/InjectionHelperPropertyHandler.kt
```diff
@@ -0,0 +1,17 @@
...
+        )
+    }
+
+    fun set(value: Variant) {
+        nodePath.set(value.unwrap())
```
CedNaru, line 15: Same here. We already know this Variant contains a NodePath so you can directly write: 
`nodePath.set(value.asNodePath()`

chippmann:
Summarizing some discussions on Discord:
@CedNaru had the idea of automating the injection by using a one shot signal that listens for `onTreeEntered`. This signal is connected inside the constructor of `Node` derived types and calls a `inject` function that calls the `JumpTableInjector` we generate.
So the user just has to use the annotation and doesn't have to call any injector in `_ready`.

So something like this:
```kotlin
//Pseudo code:
//Node.kt
class Node(): godot.Object() {
    constructor() {
        connect("ready", this, "injector", oneshot = true)
    }
    
    fun injector() = JumpTableInjector.inject(this)
}
```

chippmann:
> Summarizing some discussions on Discord:
> @CedNaru had the idea of automating the injection by using a one shot signal that listens for `onTreeEntered`. This signal is connected inside the constructor of `Node` derived types and calls a `inject` function that calls the `JumpTableInjector` we generate.
> So the user just has to use the annotation and doesn't have to call any injector in `_ready`.
> 
> So something like this:
> 
> ```kotlin
> //Pseudo code:
> //Node.kt
> class Node(): godot.Object() {
>     constructor() {
>         connect("ready", this, "injector", oneshot = true)
>     }
>     
>     fun injector() = JumpTableInjector.inject(this)
> }
> ```

I just realized: Although this works when implemented directly inside the binding, this cannot be achieved when implemented as an extension. As an extension we cannot alter the code from the base bindings.
So the question is: Is this feature useful enough to be directly implemented in the bindings, or do we have to find another way of implementing this so it works as an extension.

Ultimately it would be awesome when an extension would be able to change every part of the binding but I'm not sure how we could achieve this, nor how we would handle conflicts between extensions.

PR #165, opened 2020-07-21
Fix/cores

CedNaru:
Replace the wrong Godot calls left in the cores.
Replaced all Int/Long, Float/Double with NaturalT/RealT. If we wish to switch to another default size, we just need to change the typealias in the typeSize.kit file
The only Integers left are the one used for stuff like arrayIndex or size of containers.
Some functions return explicit sizes too, I didn't replace them. (Example: Color.toABGR64())


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
```diff
@@ -2,41 +2,51 @@
...
-import godot.gdnative.*
+import godot.gdnative.godot_string
 import kotlinx.cinterop.*
 
 
 //From Godot to Kotlin
-fun String(mem: COpaquePointer): String {
-    val gdString = mem.reinterpret<godot_string>()
-    val s = gdString.pointed.readValue().toKString()
-    godot_string_destroy(gdString)
```
chippmann, line 13: But don't we need to destroy it after readValue (which is a copy right?) to not leak memory?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
```diff
@@ -2,41 +2,51 @@
...
 }
 
 internal fun CValue<godot_string>.toKString(): String {
-    godot_string_utf8(this).useContents {
-        val ret = godot_char_string_get_data(this.ptr)?.toKString()
+    return memScoped {
+        val charString = checkNotNull(Godot.gdnative.godot_string_utf8)(this@toKString.ptr)
+        val ret = checkNotNull(Godot.gdnative.godot_char_string_get_data)(charString.ptr)?.toKString()
             ?: throw NullPointerException("Failed to convert Godot-string to Kotlin-string")
-        godot_char_string_destroy(this.ptr)
```
chippmann, line 21: Same here

godot-kotlin/godot-library/src/nativeInternal/kotlin/godot/internal/type/typeSize.kt
```diff
@@ -1,23 +1,50 @@
...
+internal inline fun Long.toNaturalTT(): NaturalT = this
+
+@PublishedApi
+internal inline fun Number.toGodotNatural(): Int = this.toInt()
+@PublishedApi
+internal inline fun Int.toGodotNatural(): Int = this
+
+//Because of float precision, it's hard to obtain two perfectly equal real numbers.
+// They are considered equal if the difference is smaller than CMD_EPSILON
+const val CMP_EPSILON: RealT = 0.00001
```
chippmann: CMD or CMP? ^^

chippmann (changes requested):

godot-kotlin/godot-library/src/nativeInternal/kotlin/godot/internal/type/typeSize.kt
CedNaru: CMP ^^

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
CedNaru: I don't know if I'm destroying a String still in use in Godot. For now I prefer to risk a leak rather than a crash.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
CedNaru: same here too

chippmann (approved):

PR #166, opened 2020-07-22
Feature/check not null

CedNaru:
Replaced CheckNotNull by a nullSafe inline function.
It just cast a nullable type to a non nullable one with "!!"
CheckNotNull is not used anymore because it's not necessary to check for null gdnative function after the initial setup as it's not mutable but just "lateinit" (not really but you get what I mean)

the content of nullSafe can still be easily be replaced by CheckNotNull in case it was really necessary so we can adapt the whole code by changing one line.

### DIFF
--- a/godot-kotlin/godot-library/src/nativeInternal/kotlin/godot/internal/type/nullSafety.kt
+++ b/godot-kotlin/godot-library/src/nativeInternal/kotlin/godot/internal/type/nullSafety.kt
@@ -1,0 +1,5 @@
+package godot.internal.type
+
+inline fun <reified T: Any> notNull(obj: T?) : T{
+    return obj!!
+}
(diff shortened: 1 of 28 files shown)


godot-kotlin/godot-library/src/nativeInternal/kotlin/godot/internal/type/nullSafety.kt
```diff
@@ -0,0 +1,5 @@
+package godot.internal.type
+
+inline fun <reified T: Any> nullSafe(obj: T?) : T{
```
raniejade:
```suggestion
inline fun <reified T: Any> nullSafe(obj: T?) : T {
```

raniejade (changes requested):

raniejade (approved):

chippmann (approved):

PR #170, opened 2020-07-23
Fix array, enumArray and enumFlag registration

chippmann:
When checking the array registration i realized we registered arrays the wrong way.
We registered them like in gdScript: `Array,Array,in`
But correctly according to the parsing logic for array hint strings: https://github.com/godotengine/godot/blob/00949f0c5fcc6a4f8382a4a97d5591fd9ec380f8/editor/editor_properties_array_dict.cpp#L546
~~it should be `19:2:19`~~ it should be `19:2` where the numbers represent the `Variant.Type.ordinal`
the above example would be a two dimensional int array: `19 = Array`, `2 = Int`

Like this the arrays are now correctly registered.

~~Not included in this PR is the fix for EnumArrays.~~ -> is now included as well

For documentation purposes (primarily for myself), the unit test i wrote to find out how these hint strings have to be formatted:
```kotlin
class Testing {

    enum class PropertyHint {
        PROPERTY_HINT_NONE, ///< no hint provided.
        PROPERTY_HINT_RANGE, ///< hint_text = "min,max,step,slider; //slider is optional"
        PROPERTY_HINT_EXP_RANGE, ///< hint_text = "min,max,step", exponential edit
        PROPERTY_HINT_ENUM, ///< hint_text= "val1,val2,val3,etc"
        PROPERTY_HINT_EXP_EASING, /// exponential easing function (Math::ease) use "attenuation" hint string to revert (flip h), "full" to also include in/out. (ie: "attenuation,inout")
        PROPERTY_HINT_LENGTH, ///< hint_text= "length" (as integer)
        PROPERTY_HINT_SPRITE_FRAME, // FIXME: Obsolete: drop whenever we can break compat. Keeping now for GDNative compat.
        PROPERTY_HINT_KEY_ACCEL, ///< hint_text= "length" (as integer)
        PROPERTY_HINT_FLAGS, ///< hint_text= "flag1,flag2,etc" (as bit flags)
        PROPERTY_HINT_LAYERS_2D_RENDER,
        PROPERTY_HINT_LAYERS_2D_PHYSICS,
        PROPERTY_HINT_LAYERS_3D_RENDER,
        PROPERTY_HINT_LAYERS_3D_PHYSICS,
        PROPERTY_HINT_FILE, ///< a file path must be passed, hint_text (optionally) is a filter "*.png,*.wav,*.doc,"
        PROPERTY_HINT_DIR, ///< a directory path must be passed
        PROPERTY_HINT_GLOBAL_FILE, ///< a file path must be passed, hint_text (optionally) is a filter "*.png,*.wav,*.doc,"
        PROPERTY_HINT_GLOBAL_DIR, ///< a directory path must be passed
        PROPERTY_HINT_RESOURCE_TYPE, ///< a resource object type
        PROPERTY_HINT_MULTILINE_TEXT, ///< used for string properties that can contain multiple lines
        PROPERTY_HINT_PLACEHOLDER_TEXT, ///< used to set a placeholder text for string properties
        PROPERTY_HINT_COLOR_NO_ALPHA, ///< used for ignoring alpha component when editing a color
        PROPERTY_HINT_IMAGE_COMPRESS_LOSSY,
        PROPERTY_HINT_IMAGE_COMPRESS_LOSSLESS,
        PROPERTY_HINT_OBJECT_ID,
        PROPERTY_HINT_TYPE_STRING, ///< a type string, the hint is the base type to choose
        PROPERTY_HINT_NODE_PATH_TO_EDITED_NODE, ///< so something else can provide this (used in scripts)
        PROPERTY_HINT_METHOD_OF_VARIANT_TYPE, ///< a method of a type
        PROPERTY_HINT_METHOD_OF_BASE_TYPE, ///< a method of a base type
        PROPERTY_HINT_METHOD_OF_INSTANCE, ///< a method of an instance
        PROPERTY_HINT_METHOD_OF_SCRIPT, ///< a method of a script & base
        PROPERTY_HINT_PROPERTY_OF_VARIANT_TYPE, ///< a property of a type
        PROPERTY_HINT_PROPERTY_OF_BASE_TYPE, ///< a property of a base type
        PROPERTY_HINT_PROPERTY_OF_INSTANCE, ///< a property of an instance
        PROPERTY_HINT_PROPERTY_OF_SCRIPT, ///< a property of a script & base
        PROPERTY_HINT_OBJECT_TOO_BIG, ///< object is too big to send
        PROPERTY_HINT_NODE_PATH_VALID_TYPES,
        PROPERTY_HINT_SAVE_FILE, ///< a file path must be passed, hint_text (optionally) is a filter "*.png,*.wav,*.doc,". This opens a save dialog
        PROPERTY_HINT_MAX,
        // When updating PropertyHint, also sync the hardcoded list in VisualScriptEditorVariableEdit
    }

    enum class VariantType {
        NIL,

        // atomic types
        BOOL,
        INT,
        REAL,
        STRING,

        // math types

        VECTOR2, // 5
        RECT2,
        VECTOR3,
        TRANSFORM2D,
        PLANE,
        QUAT, // 10
        AABB,
        BASIS,
        TRANSFORM,

        // misc types
        COLOR,
        NODE_PATH, // 15
        _RID,
        OBJECT,
        DICTIONARY,
        ARRAY,

        // arrays
        POOL_BYTE_ARRAY, // 20
        POOL_INT_ARRAY,
        POOL_REAL_ARRAY,
        POOL_STRING_ARRAY,
        POOL_VECTOR2_ARRAY,
        POOL_VECTOR3_ARRAY, // 25
        POOL_COLOR_ARRAY,

        VARIANT_MAX
    }

    @Test
    fun `Test to verify hint string parsing of this stupid godot array hint string parsing function... Grrrr`() {
        val hintStringToTest = "2/3:Enum1,Enum2"

        var subType = ""
        var subTypeHint = ""
        var subTypeHintString = ""

        val hintSubtypeSeparator = hintStringToTest.indexOf(":")

        if (hintSubtypeSeparator != -1) {
            var subTypeString = hintStringToTest.godotSubstring(0, hintSubtypeSeparator)
            val slashPosition = subTypeString.indexOf("/")

            if (slashPosition != -1) {
                subTypeHint = subTypeString.godotSubstring(slashPosition + 1, subTypeString.length - slashPosition - 1)
                subTypeString = subTypeString.godotSubstring(0, slashPosition)
            }

            subTypeHintString = hintStringToTest.godotSubstring(hintSubtypeSeparator + 1, hintStringToTest.length - hintSubtypeSeparator - 1)
            subType = subTypeString
        }

        assertEquals(VariantType.INT.ordinal, subType.toInt())
        assertEquals(PropertyHint.PROPERTY_HINT_ENUM.ordinal, subTypeHint.toInt())
        assertEquals("Enum1,Enum2", subTypeHintString)
    }
}

fun String.godotSubstring(from: Int, chars: Int): String {
    val p_from = from
    var p_chars = chars

    if (p_chars == -1)
        p_chars = length - p_from;

    if (isEmpty() || p_from < 0 || p_from >= length || p_chars <= 0)
        return "";

    if ((p_from + p_chars) > length) {
        p_chars = length - p_from;
    }

    if (p_from == 0 && p_chars >= length) {
        return this
    }

    return substring(p_from, p_from + p_chars)
}
```


raniejade (approved):
Gotta love how gdnative is so undocumented 😂

chippmann:
It now also fixes enumFlag registration (thanks @CedNaru!)

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/classBuilderDsl.kt
```diff
@@ -158,38 +158,37 @@ class ClassBuilder<T : Object> internal constructor(val classHandle: ClassHandle
...
         rpcMode: RPCMode
     ) {
-        val variantArray = IntVariantArray()
-        if (default != null) {
-            default.asVariantArray().forEach {
-                variantArray.append((it as K).ordinal.toNaturalT())
-            }
+        var intFlag = 0
+        default?.forEach { enum ->
+            intFlag += 1 shl enum.ordinal
```
CedNaru, line 179: We should add security for cases when the users try to register an Enum with more than 32 entries.
We are using one bit for each entry, if we go above 32 we are just going to break the final Int.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/properties.kt
```diff
@@ -34,19 +34,36 @@ class MutableEnumPropertyHandler<T : Object, R : Enum<R>>(
...
-            .toSet()
-        property.set(instance, transformedValue)
+        val intFlag = value.asInt()
+
+        val enums = mutableSetOf<R>()
+        var bit = 1
+
+        for (i in 0 until Int.SIZE_BITS) {
+            if ((intFlag and bit) > 0) {
+                val element = converter(i.toNaturalT())
```
CedNaru: Do we really need a toNaturalT here? We are dealing with enum ordinals, it should always be a Int anyway.

raniejade (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/classBuilderDsl.kt
chippmann: Hmm yeah you are right. If it's ok for you I'll do a separate issue for this.

chippmann: https://github.com/utopia-rise/godot-kotlin/issues/175

PR #171, opened 2020-07-24
Fix/global functions

### LINKED ISSUE #168: [Master-Merge ToDo] Improve GD.print and fix frozen issue
Currently if one just wants to print a string, one has to write:
`GD.print(arrayOf(Variant("Property: ${this::rangeInt.name} -> $rangeInt")))` which is very convoluted

I propose a signature like this:
`fun print(vararg args: Any)`

CedNaru:
The tyranny of the single StringBuilder is over.

Resolves #168

### DIFF
--- a/godot-kotlin/godot-library/src/nativePublic/kotlin/godot/global/GDPrint.kt
+++ b/godot-kotlin/godot-library/src/nativePublic/kotlin/godot/global/GDPrint.kt
@@ -2,37 +2,38 @@ package godot.global
 
 import godot.core.Godot
 import godot.core.Variant
+import kotlin.native.concurrent.AtomicReference
 
 internal interface GDPrint {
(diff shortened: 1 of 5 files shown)


godot-kotlin/godot-library/src/nativePublic/kotlin/godot/global/GDPrint.kt
```diff
@@ -2,37 +2,40 @@ package godot.global
 
 import godot.core.Godot
 import godot.core.Variant
+import kotlin.native.concurrent.AtomicReference
```
chippmann: unused import

chippmann (changes requested):

chippmann:
Resolves #168

chippmann (approved):

PR #178, opened 2020-07-26
Add initial documentation

raniejade:
This is a low effort port of the docs of my binding. To test it, you must have `pip` installed and run `docs/run.sh`. `build.sh` will only be used in CI.


CedNaru (approved):

docs/src/doc/index.md
```diff
@@ -0,0 +1,10 @@
+[![GitHub](https://img.shields.io/github/license/utopia-rise/godot-kotlin?style=flat-square)](LICENSE)
+[![GitHub Workflow Status](https://img.shields.io/github/workflow/status/raniejade/godot-kotlin/CI?style=flat-square)](https://github.com/utopia-rise/godot-kotlin/actions?query=workflow%3ACI)
```
chippmann: Wrong link

docs/src/doc/setup/gradle.md
```diff
@@ -0,0 +1,94 @@
...
+    mavenCentral()
+}
+
+godot {
+    // Build a debug binary
+    debug.set(true)
+
+    configureTargets {
+        compilations.getByName("main") {
+            cinterops {
```
chippmann: I would remove this in the initial getting started as its not needed and it might confuse people

docs/src/doc/setup/gradle.md
```diff
@@ -0,0 +1,94 @@
...
+```kotlin
+import godot.*
+import godot.core.*
+
+@RegisterClass
+class Simple: Spatial() {
+
+    @RegisterFunction
+    override fun _ready() {
+        gprint("Hello Godot from Kotlin!")
```
chippmann: Wrong print function

docs/src/doc/user-guide/methods.md
```diff
@@ -0,0 +1,17 @@
...
+`Variant`, additionally the function must be annotated with `@RegisterFunction`. This binding only support 
+methods with at most 10 parameters at the moment.
+
+```kotlin
+@RegisterClass
+class RotatingCube: Spatial() {
+    
+    @RegisterFunction  
+    override fun _ready() {
+        gprint("I am ready!")
```
chippmann: Wrong print

docs/src/doc/user-guide/methods.md
```diff
@@ -0,0 +1,17 @@
...
+    }
+}
+```
+
+## Virtual methods
+Virtual methods (like `_ready`, `_process` and `_physics_process`) are declared as overridable methods. The default implementation throws a `NotImplementedException`, so you have to override it if you plan to expose a virtual method to Godot.
```
chippmann, line 17: Maybe add here that they also need the @registerFunction annotation

chippmann (changes requested):
Very cool. Even though "low effort" some comments about some incorrectnesses which i think needs fixing from the start.
But very neat!

docs/src/doc/user-guide/methods.md
raniejade: It's already mentioned on the same page, I don't think we need to repeat that info here.

chippmann: fair point

chippmann (approved):

piiertho (approved):

PR #181, opened 2020-07-26
reinterpret retvar as godot_string and then convert it to Kotlin String in icalls returning String

piiertho:
We were trying to convert a pointer to KString, this fixes this behaviour by first reinterpret pointer as `godot_string`, then read its value and anding by converting value to Kotlin String.


CedNaru:
I'm a bit confused, isn't Godot supposed to return us a Variant in any case, not directly a Godot_string ?

raniejade:
Yeah, I thought it was a `godot_variant` not a `godot_string`. Can you confirm @piiertho?

piiertho:
I have a better way to do, will be fixed with rework of allocations in icalls.

PR #186, opened 2020-07-27
[Entry-Generation] Replace `Variant.wrap` calls for registered properties and functions

chippmann:
This replaces all `Variant.wrap` calls for all registered functions and properties.
~~Signals are not touched yet. Depending on my time today and when you guys do the review i either include it in this PR or in a separate.~~ Do we need to do anything for the signals? we only have one wrap function there:
```kotlin
@PublishedApi
    internal fun connect(
        instance: Object,
        target: Object,
        method: String,
        binds: List<Any>?,
        flags: Long
    ) {
        val extraArgs = VariantArray()
        binds?.forEach { extraArgs.append(Variant.wrap(it)) }
        instance.connect(name, target, method, extraArgs, flags)
    }
```

but we cannot change that as we don't know the type of the binds right?

### DIFF
--- a/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/FunctionRegistrationGenerator.kt
+++ b/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/FunctionRegistrationGenerator.kt
@@ -4,16 +4,21 @@ import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.MemberName.Companion.member
 import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.asTypeName
 import godot.entrygenerator.extension.getAnnotationValue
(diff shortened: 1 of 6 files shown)


CedNaru (approved):

piiertho:
> This replaces all `Variant.wrap` calls for all registered functions and properties.
> ~Signals are not touched yet. Depending on my time today and when you guys do the review i either include it in this PR or in a separate.~ Do we need to do anything for the signals? we only have one wrap function there:
> 
> ```kotlin
> @PublishedApi
>     internal fun connect(
>         instance: Object,
>         target: Object,
>         method: String,
>         binds: List<Any>?,
>         flags: Long
>     ) {
>         val extraArgs = VariantArray()
>         binds?.forEach { extraArgs.append(Variant.wrap(it)) }
>         instance.connect(name, target, method, extraArgs, flags)
>     }
> ```
> 
> but we cannot change that as we don't know the type of the binds right?

I think we cannot do anything about it as they are kind of variadic :/

piiertho (approved):

PR #197, opened 2020-07-30
Add ResourceLoader.load<T> and Node.getNode<T> extensions

raniejade:
Instead of writing:

```kotlin
(getNode(NodePath(spriteNodePath)) as Sprite).texture = textureForSprite
```

You can do:

```kotlin
getNode<Sprite>(pathAsString).texture = textureForSprite
```


chippmann (approved):

CedNaru:
I think it's weird to have a global GD.load(String) function in the public singleton but a generic \<T\> load(String) as an extension of Resource loader.
We should either implement it as a global function or just delete the GDResource directly for consistency.

raniejade:
@CedNaru I forgot that `GD.load` exists, I can remove `ResourceLoader.load` extension. People are probably more familiar with `GD.load` rather than using `ResourceLoader`.

CedNaru:
Okay let's do that.

CedNaru (approved):

Issue #157, opened 2020-07-15
[Master-Merge ToDo] Signal connection methods don't receive the right functions as params


chippmann:
Compiles when it shouldn't:
![success](https://user-images.githubusercontent.com/22662033/87545120-c7137280-c6a7-11ea-844e-773a3c4d6fe2.png)
Compiles not when it should:
![error](https://user-images.githubusercontent.com/22662033/87545114-c549af00-c6a7-11ea-9306-d2116372e9b4.png)

chippmann:
The function reolves to `    public final inline fun <A0, reified K : () -> kotlin.Unit> godot.core.Signal1<A0>.connect(target: godot.Object, method: K, binds: kotlin.collections.List<kotlin.Any>? /* = compiled code */, flags: kotlin.Long /* = compiled code */): kotlin.Unit { /* compiled code */ }
`
inside the `Object` class

Which is this generated function:
```kotlin
  inline fun <A0, reified K : () -> Unit> Signal1<A0>.connect(
    target: Object,
    method: K,
    binds: List<Any>? = null,
    flags: Long = 0
  ) {
    val methodName = (method as KCallable<Unit>).name
    connect(this@Object, target, methodName, binds, flags)
  }
```

I guess `A0` needs to be inserted before `Unit`:
```kotlin
val methodName = (method as KCallable<A0, Unit>).name
```
and `A0` needs to be added in `<A0, reified K : () -> kotlin.Unit>` to be `<A0, reified K : (A0) -> kotlin.Unit>`

chippmann:
> The function reolves to ` public final inline fun <A0, reified K : () -> kotlin.Unit> godot.core.Signal1<A0>.connect(target: godot.Object, method: K, binds: kotlin.collections.List<kotlin.Any>? /* = compiled code */, flags: kotlin.Long /* = compiled code */): kotlin.Unit { /* compiled code */ }`
> inside the `Object` class
> 
> Which is this generated function:
> 
> ```kotlin
>   inline fun <A0, reified K : () -> Unit> Signal1<A0>.connect(
>     target: Object,
>     method: K,
>     binds: List<Any>? = null,
>     flags: Long = 0
>   ) {
>     val methodName = (method as KCallable<Unit>).name
>     connect(this@Object, target, methodName, binds, flags)
>   }
> ```

I guess `A0` needs to be inserted before `Unit`:
```kotlin
val methodName = (method as KCallable<A0, Unit>).name
```

PR #198, opened 2020-07-30
Fix various icall crashes

raniejade:


buildSrc/src/main/kotlin/godot/codegen/ICall.kt
```diff
@@ -170,13 +170,13 @@ class ICall(
                         val typeManagerClass = ClassName("godot.core", "TypeManager")
                         if (returnTypeClassSimpleName != "Object") {
                             codeBlockBuilder.add(
-                                "    %M(retVar) as %T\n",
+                                "    %M(retVar.pointed.value!!) as %T\n",
```
raniejade, line 173: previously `wrap` was given a `void**` instead of a `void*`

buildSrc/src/main/kotlin/godot/codegen/generationEntry.kt
```diff
@@ -20,7 +20,7 @@ infix fun File.generateApiFrom(jsonSource: File) {
     val iCallFileSpec = FileSpec
         .builder("godot.icalls", "__icalls")
         .addFunction(generateICallsVarargsFunction())
-        .addImport("kotlinx.cinterop", "set", "get")
+        .addImport("kotlinx.cinterop", "set", "get", "pointed")
```
raniejade, line 23: pointed is an extension method, I have to import it manually.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -68,14 +67,15 @@ object Godot {
             )
             languageIndexRef.compareAndSet(languageIndexRef.value, index)
         }
+        TypeManager.registerEngineTypes()
```
raniejade, line 70: `TypeManager.registerEngineTypes()` should be called here instead of `init` as it depends on `languageIndex`.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/Godot.kt
```diff
@@ -68,14 +67,15 @@ object Godot {
...
     }
 
     fun nativescriptTerminate(handle: COpaquePointer) {
+        TypeManager.dispose()
```
raniejade, line 74: same as above

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/TypeManager.kt
```diff
@@ -67,6 +64,7 @@ internal object TypeManager {
         val copy = mutableListOf<COpaquePointer>()
         copy.addAll(tags.value)
         copy.add(tag)
+        tags.compareAndSet(tags.value, copy.freeze())
```
raniejade, line 67: we were not tracking the stable refs properly before.

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
```diff
@@ -37,11 +37,11 @@ class TestingClass : Node() {
     @RegisterSignal
     val signal2 by signal<Boolean, String>("param1", "param2")
 
-    init {
-        signal0.connect(this, this::signal0Receiver)
-        signal1.connect(this, this::signal1Receiver)
-        signal2.connect(this, this::signal2Receiver)
-    }
+//    init {
```
raniejade: this is a no-no, `this` is not initialized yet here. We need to avoid the init block.

piiertho (approved):
Thanks for the bug fix !

samples/mini-games/src/godotMain/kotlin/example/TestingClass.kt
chippmann: True. This was really a bad testing sample that was only there to check the fix for #157. Anyway we still need to add that bit to the documentation as preventing it is not possible.

chippmann (approved):

PR #199, opened 2020-07-31
Enhancement/gdstring as inline class

piiertho:
This transforms typealias GdString to inline class, to simplify use of Godot Strings.
This convert String arguments of icalls to GdString before passing them to godot, and then destroy them when not used anymore.

### DIFF
--- a/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
+++ b/godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
@@ -7,52 +7,52 @@ import godot.gdnative.godot_string_layout
 import godot.internal.type.nullSafe
 import kotlinx.cinterop.*
 
-typealias GdString = CValue<godot_string_layout>
+inline class GdString(val value: CValue<godot_string_layout>) {
 
-//From Godot to Kotlin
-internal fun String(ptr: COpaquePointer) = String(ptr.reinterpret())
-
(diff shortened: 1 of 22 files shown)


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/String.kt
```diff
@@ -7,52 +7,52 @@ import godot.gdnative.godot_string_layout
...
-    return memScoped {
-        val charString = nullSafe(Godot.gdnative.godot_string_utf8)(this@toKString.ptr)
-        val ret = nullSafe(Godot.gdnative.godot_char_string_get_data)(charString.ptr)?.toKString()
-            ?: throw NullPointerException("Failed to convert Godot-string to Kotlin-string")
-        nullSafe(Godot.gdnative.godot_char_string_destroy)(charString.ptr)
-        nullSafe(Godot.gdnative.godot_string_destroy)(this@toKString.ptr)
-        ret
-    }
-}
+internal fun String(ptr: CPointer<godot_string>) = GdString(ptr.pointed.readValue())
```
raniejade: `String` -> `GDString`, this can cause confusion between the Kotlin's `String` class.

raniejade (changes requested):

raniejade (approved):

raniejade:
I've tested this locally, merging. I really want to send the 3D platformer PR :P

PR #203, opened 2020-08-02
Add init of pointer for generated singletons

piiertho:
Generated singletons were not generated properly.
This adds an init block to them to initialize their pointers.


raniejade (approved):

raniejade:
@piiertho I'll merge this now, I need it for my branch.

PR #204, opened 2020-08-02
Add 3D Platformer demo

raniejade:


buildSrc/src/main/kotlin/godot/codegen/Class.kt
```diff
@@ -16,6 +16,8 @@ class Class @JsonCreator constructor(
     var baseClass: String,
     @JsonProperty("singleton")
     val isSingleton: Boolean,
+    @JsonProperty("singleton_name")
+    val singletonName: String,
```
raniejade, line 20: For singletons, we need to use this property instead of the class name. i.e `_ClassDB` which has a singleton name of `ClassDB`.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -762,6 +795,72 @@ class Basis(
...
+
+
+    fun set(xAxis: Vector3, yAxis: Vector3, zAxis: Vector3) {
+        setAxis(0, xAxis)
+        setAxis(1, yAxis)
+        setAxis(2, zAxis)
+    }
+
+    /*
+     * GDScript related members
```
raniejade, line 813: Everything below here are only available in GDScript, i.e meaning the core cpp implementation does not have them declared.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -762,6 +795,72 @@ class Basis(
...
+     */
+    constructor(xAxis: Vector3, yAxis: Vector3, zAxis: Vector3) : this() {
```
raniejade, line 815: Basis is exposed as column major in GDScript but internally uses a row major format

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform.kt
```diff
@@ -65,7 +65,7 @@ class Transform(
 
     //CONSTRUCTOR
     constructor() :
-        this(Basis(1, 0, 0, 0, 1, 0, 0, 0, 1), Vector3(0, 0, 0))
+        this(Basis(), Vector3(0, 0, 0))
```
raniejade, line 68: default constructor for `Basis` is set as IDENTITY.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Transform.kt
```diff
@@ -280,9 +280,9 @@ class Transform(
      */
     fun xform(vector: Vector3): Vector3 =
         Vector3(
-            _basis.x.dot(vector) + _origin.x,
-            _basis.y.dot(vector) + _origin.y,
-            _basis.z.dot(vector) + _origin.z
+            _basis._x.dot(vector) + _origin.x,
```
raniejade, line 283: `x`, `y`, and `z` are only visible in GDScript - so therefore they shouldn't be used internally.

samples/3d-platformer/3d-platformer.gdnlib
```diff
@@ -0,0 +1,14 @@
+[entry]
+
+X11.64="res://build/bin/linuxX64/debugShared/lib3d_platformer.so"
```
chippmann, line 3: IMO we should add the ones for windows and macOs as well right away

samples/3d-platformer/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+plugins {
+    kotlin("multiplatform") version "1.3.71"
```
chippmann: why not 72?

samples/3d-platformer/src/godotMain/kotlin/Enemy.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        lv -= deaccelDir * deaccelDir.dot(lv) + deaccelDir * dspeed
+
+        state.linearVelocity = lv
+        prevAdvance = advance
+    }
+
+    @RegisterFunction
+    fun _die() {
```
chippmann, line 111: why prefixed with `_`?

chippmann (changes requested):
Also Code_Owner is not set up.
Apart from that; very good work!

samples/3d-platformer/src/godotMain/kotlin/Enemy.kt
raniejade: I'm staying true to the original material, so I didn't change the method name.

samples/3d-platformer/build.gradle.kts
raniejade: same as mini-mages, but I'll try updating.

samples/3d-platformer/3d-platformer.gdnlib
raniejade: Ahh, this is because of me only running `buildLinuxX64` locally. I'll add the others.

CedNaru (approved):

buildSrc/src/main/kotlin/godot/codegen/Class.kt
```diff
@@ -222,8 +224,8 @@ class Class @JsonCreator constructor(
             typeBuilder.addInitializerBlock(
                 CodeBlock.of("""
                             |%M {
-                            |    val ptr = %M(%T.gdnative.godot_global_get_singleton).%M("$oldName".%M.ptr)
-                            |    %M(ptr) { "No instance found for singleton $newName" }
+                            |    val ptr = %M(%T.gdnative.godot_global_get_singleton).%M("$singletonName".%M.ptr)
+                            |    %M(ptr) { "No instance found for singleton $singletonName" }
```
piiertho, line 228: I would continue to use newName in log, so that it is easier to retrieve from where comes the error.

samples/3d-platformer/enemy.gd
```diff
@@ -0,0 +1,81 @@
+
+extends RigidBody
```
piiertho, line 2: Why do we have gdscript ?

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
```diff
@@ -9,64 +9,27 @@ import kotlinx.cinterop.*
...
-    var z
-        get() = Vector3(_z)
-        set(value) {
-            _z = Vector3(value)
-        }
-
-    inline fun <T> z(block: Vector3.() -> T): T {
-        return _z.block()
+    init {
+        _x[0] = 1.0
```
piiertho: We don't need to use array access operator that uses switch. We already know which one we want.

piiertho (changes requested):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
CedNaru: Heretic

buildSrc/src/main/kotlin/godot/codegen/Class.kt
raniejade: It's probably harder to debug though, as the newName != singletonName - you have to look up at the source to know what name are we exactly passing.

samples/3d-platformer/enemy.gd
raniejade: I kept them for reference.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/type/Basis.kt
raniejade: I'll sneak in more switches @CedNaru 😂 . Kidding aside, yep I'll fix that.

samples/3d-platformer/3d-platformer.gdnlib
raniejade: @chippmann This file is generated, and it is impossible to know the path of the shared libs in windows and osx from a linux machine as the paths are only calculated when a link task is executed. Maybe it's best to just ignore this file? You probably will trigger a separate build for each platform you deploy to when releasing, so the correct paths will be present.

chippmann: Yeah you are right. But i would not ignore it in the .gitignore if thats what you mean.

chippmann (approved):

raniejade:
@piiertho :ping_pong:

PR #208, opened 2020-08-04
Use unix path separator for Windows

raniejade:


samples/3d-platformer/settings.gradle.kts
```diff
@@ -1,3 +1,5 @@
+rootProject.name = "platformer3d"
```
raniejade, line 1: no need to rename the directory

piiertho (approved):

CedNaru (approved):

PR #212, opened 2020-08-04
Improve documentation

chippmann:


docs/mkdocs.yml
```diff
@@ -36,3 +36,7 @@ nav:
     - Methods: user-guide/methods.md
     - Signals: user-guide/signals.md
     - Properties: user-guide/properties.md
+  - Contribution:
+    - General: contribution/general.md
+    - Decisions / Guides:
```
raniejade: I don't think this should be here, just a simple `design-docs` folder on the root repo should be fine. No need to expose it.

raniejade (changes requested):

docs/mkdocs.yml
chippmann: Agree. Updated it

CedNaru (approved):

raniejade (approved):

PR #215, opened 2020-08-05
Make samples a composite build

raniejade:
Made some bug fixes to the gradle plugin that surfaced when converting it to the a composite build. This also adds github workflows for the two samples.


plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotExtension.kt
```diff
@@ -13,14 +13,24 @@ open class GodotExtension(objects: ObjectFactory) {
...
     internal val entrySourceDir = objects.directoryProperty()
 
     private var configureTargets: KotlinNativeTarget.() -> Unit = {}
 
     fun platforms(vararg platforms: Platform) {
-        this.platforms.set(platforms.toList())
+        configure(platforms.toList())
+    }
+
+    fun defaultPlatforms() {
```
raniejade, line 26: This was supposed to be in `1.4-rc` but I needed it here to fix the gradle plugin.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -44,48 +38,47 @@ class GodotPlugin : Plugin<Project> {
     }
 
     private fun setupKotlinPlugin(project: Project, mpp: KotlinMultiplatformExtension, godot: GodotExtension) {
-        project.afterEvaluate {
```
raniejade, line 47: `afterEvaluate` is not really compatible with the multiplatform plugin.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -131,13 +125,14 @@ class GodotPlugin : Plugin<Project> {
                             NativeBuildType.RELEASE
                         }
                         sharedLib(buildTypes = listOf(buildType)) {
+                            generateGdnlibTask.dependsOn(linkTask)
                             // this will create a task dependency
                             // build -> build<Target> -> link<BuildType><Target>
-                            project.tasks.register("build${target.name.capitalize()}") {
+                            project.tasks.create("build${target.name.capitalize()}") {
```
raniejade, line 130: Use create instead of register otherwise `./gradlew build` won't work.

plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/subplugin/GodotSubPlugin.kt
```diff
@@ -52,7 +53,7 @@ class GodotSubPlugin : KotlinGradleSubplugin<AbstractCompile> {
         //&& task is KotlinNativeCompile -> we cannot do this check here as the compiler plugin get's called in the
         // linking stage again, which then fails as the subPlugin is disabled for it and thus does not provide the
         // needed command line arguments
-        project.plugins.hasPlugin(GodotPlugin::class.java)
+        project.extensions.findByType<GodotExtension>() != null
```
raniejade, line 56: I totally forgot about this, always check for the extension not the plugin class. This can cause issue when you are checking for a presence of your plugin in the configuration phase.

samples/3d-platformer/platformer3d.gdnlib
```diff
@@ -1,10 +1,14 @@
 [entry]
 
+X11.64="res://build/bin/linuxX64/debugShared/libplatformer3d.so"
```
raniejade, line 3: With my fix, the gdnlib now properly contains all the paths!

.github/workflows/check-pr-samples-3d-platformer.yaml
```diff
@@ -0,0 +1,40 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            task: buildLinuxX64
+          - os: macos-latest
+            task: -x linkDebugSharedLinuxX64 buildMacosX64
+          - os: windows-latest
+            task: -x linkDebugSharedLinuxX64 buildWindowsX64
```
chippmann, line 27: Why isd the link step for linux necessary for windows and macOS?

.github/workflows/check-pr-samples-mini-games.yaml
```diff
@@ -0,0 +1,40 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            task: buildLinuxX64
+          - os: macos-latest
+            task: -x linkDebugSharedLinuxX64 buildMacosX64
+          - os: windows-latest
+            task: -x linkDebugSharedLinuxX64 buildWindowsX64
```
chippmann, line 27: same here?

.github/workflows/check-pr-samples-mini-games.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v2
+      - name: Checkout submodules
+        uses: snickerbockers/submodules-init@v4
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: samples/3d-platformer
```
chippmann: wrong path

samples/3d-platformer/build.gradle.kts
```diff
@@ -4,11 +4,11 @@ plugins {
 }
 
 repositories {
-    mavenLocal()
```
chippmann, line 7: Why no mavenLocal?

samples/3d-platformer/build.gradle.kts
```diff
@@ -4,11 +4,11 @@ plugins {
...
     jcenter()
 }
 
 
 godot {
     debug.set(true)
+    defaultPlatforms()
```
chippmann, line 13: I also suggest we change the name of our `Platform` enum to `GodotPlatform` as the IDE has trouble importing the correct one and you always have to import the correct ones yourself

samples/mini-games/build.gradle.kts
```diff
@@ -6,9 +6,7 @@ buildscript {
 }
 
 repositories {
-    mavenLocal()
```
chippmann, line 9: why no mavenLocal?

chippmann (changes requested):

samples/3d-platformer/build.gradle.kts
raniejade: There's no need for it.

.github/workflows/check-pr-samples-3d-platformer.yaml
raniejade: It's excluding the `linkDebugSharedLinuxX64 `, unfortunately `buildMacosX64` also triggers the link commands for linux. Also do note you can build linux binaries in macos and windows - if I don't exclude them the build lasts for 20mins+!

samples/3d-platformer/build.gradle.kts
raniejade: Let's do that in a separate PR, not here.

.github/workflows/check-pr-samples-3d-platformer.yaml
chippmann: ah sorry didn't pay attention to the `-x`

samples/3d-platformer/build.gradle.kts
chippmann: hmm but it would be if it were a separate project. If i understand your composite build setup correctly, the `build.gradle.kts` (but not `settings.gradle.kts` but there it's commented) from the samples files look the same regardless if we are using composite builds or not. So i would leave them as they are even if they contain unused or redundant parts as they serve the purpose of being examples of how you can setup a project.
Or we can comment it.

That said, I'm nitpicking here. I'm fine either way :-)

samples/3d-platformer/build.gradle.kts
chippmann: :+1:

chippmann (approved):

samples/3d-platformer/build.gradle.kts
raniejade: Even without composite build, `mavenLocal` is not needed because we don't expect users to build the library first - but used the either the actual release or the dev release.

CedNaru (approved):
I have no idea what I'm approving.

piiertho (approved):

PR #216, opened 2020-08-07
Register functions, properties and signals snake case

### LINKED ISSUE #141: Register functions with GDScript friendly names
@2shady4u described the need of defining GDScript friendly method names while registering classes here: https://github.com/utopia-rise/godot-kotlin/issues/66

This is a good idea as sometimes you want to register a method with godot and access it through kotlin as well. With the current solution, you either declare your function name kotlin style camelCase or GDScript style snake_case. But in one of the two languages you end up calling a function which is not written in the correct casing.

If we provide something like this:
```kotlin
@RegisterFunction(rpcMode = RPCMode.None, registeredName = "my_cool_function")
fun myCoolFunction(){}
```
one can call it from kotlin with `myCoolFunction` and from GDScript with `my_cool_function`. 

This would lead to the following limitation though: when using `someNode.callDeferred("")` one would have to write the name from GDScript like this: `someNode.callDeferred("my_cool_function")` and could not use `someNode.callDeferred(SomeNode::myCoolFunction.name)`. But i guess when we say that in the docs it should be fine. Or we could provide extension functions which map the `name`to the correct casing.

Of course this should be completely optional as it would only be used if users mix languages heavily.

Once we are about to implement it, we should discuss the implementation details first.

chippmann:
Resolves #141


godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/signals.kt
```diff
@@ -6,7 +6,7 @@ abstract class Signal(
     val name: String
 ) {
     protected fun emitSignal(instance: Object, vararg args: Any?) {
-        instance.emitSignal(name, *args)
+        instance.emitSignal(name.camelToSnakeCase(), *args) //not using `camelcaseToUnderscore` to prevent a call to godot for each signal emission
```
raniejade: Instead of doing it every time we want to emit a signal we can do it once in `SignalDelegateProvider`.

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/signals.kt
```diff
@@ -17,7 +17,7 @@ abstract class Signal(
         binds: VariantArray?,
         flags: Long
     ) {
-        instance.connect(name, target, method, binds ?: VariantArray(), flags)
+        instance.connect(name.camelToSnakeCase(), target, method, binds ?: VariantArray(), flags) //not using `camelcaseToUnderscore` to prevent a call to godot for each signal connection
```
raniejade: same here

raniejade (commented):
I had a few comments but this is going to be awesome! :)

raniejade:
Also we probably need a specialized extension for `callDeferred`.

```kotlin
inline fun <reified T: KFunction<*>> Object.callDeferred(method: T, vararg args: Any?) {
    callDeferred(method.name, args)
}

class S : Node() {
    
    @RegisterFunction
    override fun _ready() {
        callDeferred(this::fooBar)
    }
    
    @RegisterFunction
    fun fooBar() {
        
    }
}
```

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/ClassHandle.kt
```diff
@@ -151,7 +151,7 @@ internal class ClassHandle<T : Object>(
             nullSafe(Godot.nativescript.godot_nativescript_register_property)(
                 nativescriptHandle,
                 className.cstr.ptr,
-                propertyName.cstr.ptr,
+                propertyName.camelToSnakeCase().cstr.ptr, //not using `camelcaseToUnderscore` to prevent a call to godot for each property
```
raniejade, line 154: what is `camelcaseToUnderscore`?

raniejade (approved):

CedNaru (approved):

godot-kotlin/godot-library/src/nativeCore/kotlin/godot/core/ClassHandle.kt
chippmann:
```kotlin
/** Convert a CamelCase string to a Snake_Case one */
fun String.camelcaseToUnderscore() = asGDString {
    GdString(nullSafe(Godot.gdnative.godot_string_camelcase_to_underscore)(it.value.ptr)).toKString()
}
```
defined inside `Util.kt`

PR #218, opened 2020-08-08
Drop enum value constants

raniejade:
This PR drops the constant values that are actually enum values. If we had the generated classes committed we could have seen a diff of the actual changes :P.


piiertho (approved):
Thx !

PR #220, opened 2020-08-08
Rename Platform to GodotPlatform and improve setup guide

raniejade:


CedNaru (approved):

docs/src/doc/setup/gradle.md
```diff
@@ -15,33 +15,67 @@ The above command will create three files, which will be empty for now.
...
+
+## Creating your first class
+
 Let's create a file `src/godotMain/kotlin/Simple.kt` with the following contents.
 
 ```kotlin
 import godot.*
-import godot.core.*
+import godot.annotation.*
+import godot.global.*
```
chippmann: `import godot.global.*` is unused but `import godot.core.GD` is missing

docs/src/doc/setup/gradle.md
```diff
@@ -55,16 +89,15 @@ class Simple: Spatial() {
 
 More will be explained in the [classes](../user-guide/classes.md) section of user guide, but for now `@RegisterClass` will register
 the annotated class to Godot.
-
-!!! note ""
-    The plugin automatically generates the appropriate `gdns` files which can be found at `src/gdns`. It is up to you whether you want to include those files in source control or not.
     
 Now we can trigger a build.
```
chippmann, line 102: IMO `org.gradle.jvmargs=-Xmx3G` inside `gradle.properties` is missing as the first build failed for me because of running out of memory

chippmann (changes requested):
A few comments otherwise awesome work!

docs/src/doc/setup/gradle.md
chippmann: maybe also an explanation why we need more memory would be good in an info box or something like that

piiertho (approved):

chippmann (approved):

PR #227, opened 2020-08-09
Add snake_case registration documentation

### LINKED ISSUE #225: Add registration documentation about snake_case
When implementing #141 I forgot to add proper documentation about the registered names.
This issue aims to resolve that.

chippmann:
Resolves #225

### DIFF
--- a/docs/src/doc/user-guide/methods.md
+++ b/docs/src/doc/user-guide/methods.md
@@ -13,6 +13,12 @@ class RotatingCube: Spatial() {
 }
 ```
 
+!!! danger ""
+    All methods that you register, are registered in `snake_case`!  
+    So `myVeryCoolFunction` will become `my_very_cool_function`  
+    This is done for easier GDScript integration and that Godot can properly call overridden virtual functions from other languages.  
+    Keep this in mind when calling kotlin functions from other languages or when using functions like `call` and not using our extension functions which handle the conversion for you!
+
 ## Virtual methods
 Virtual methods (like `_ready`, `_process` and `_physics_process`) are declared as overridable methods. The default implementation throws a `NotImplementedException`, so you have to override it if you plan to expose a virtual method to Godot.  
 Remember; just overriding is not enough to use that function. You have to explicitly register it as well with `@RegisterFunction` like you have to with your own methods.
--- a/docs/src/doc/user-guide/methods.md
+++ b/docs/src/doc/user-guide/methods.md
@@ -13,11 +13,8 @@ class RotatingCube: Spatial() {
 }
 ```
 
-!!! danger ""
-    All methods that you register, are registered in `snake_case`!  
-    So `myVeryCoolFunction` will become `my_very_cool_function`  
-    This is done for easier GDScript integration and that Godot can properly call overridden virtual functions from other languages.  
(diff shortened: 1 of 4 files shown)


raniejade (changes requested):
I find the wrapping on the lines weird, are we able fix them? There's too much unused space from the right side.

Screenshot from 2020-08-09 18-31-19

docs/src/doc/user-guide/methods.md
```diff
@@ -13,6 +13,12 @@ class RotatingCube: Spatial() {
 }
 ```
 
+!!! danger ""
+    All methods that you register, are registered in `snake_case`!  
+    So `myVeryCoolFunction` will become `my_very_cool_function`
```
raniejade: Let it wrap naturally, it's okay that the lines are longer. When rendered it will be of the appropriate length.

docs/src/doc/user-guide/properties.md
```diff
@@ -19,6 +19,12 @@ class RotatingCube: Spatial() {
 }
 ```
 
+!!! danger ""
+    All properties that you register, are registered in `snake_case`!
```
raniejade: same here

docs/src/doc/user-guide/signals.md
```diff
@@ -10,6 +10,13 @@ class RotatingCube: Spatial() {
 }
 ```
 
+!!! danger ""
+    All signals that you register, are registered in `snake_case`!
```
raniejade: likewise

docs/src/doc/user-guide/methods.md
```diff
@@ -13,6 +13,12 @@ class RotatingCube: Spatial() {
 }
 ```
 
+!!! danger ""
```
raniejade: important is probably better here - because we give an important information not a warning for certain pitfalls.

docs/src/doc/user-guide/methods.md
chippmann: I wasn't wrapping it because of line lenght, i was wrapping because i thought it makes sense. But I'll change it :-)

piiertho (approved):

raniejade (changes requested):
Just noticed that there is some outdated info in `API Differences`:

> Additionally, signals are mapped to properties of type Signal and must start with a prefix signal (check Signals section for more details). The prefix is dropped during registration, so the signal signalReverseChanged is known in Godot as reverseChanged. This is done to avoid naming conflicts with other members of a class. There is no signal type in GDScript, signals are only referenced by name so they can have the same name as methods and/or properties in the same class. 

Can you change that as well?

chippmann:
> Just noticed that there is some outdated info in `API Differences`:
> 
> > Additionally, signals are mapped to properties of type Signal and must start with a prefix signal (check Signals section for more details). The prefix is dropped during registration, so the signal signalReverseChanged is known in Godot as reverseChanged. This is done to avoid naming conflicts with other members of a class. There is no signal type in GDScript, signals are only referenced by name so they can have the same name as methods and/or properties in the same class.
> 
> Can you change that as well?

sure

raniejade (approved):

PR #232, opened 2020-08-12
Regenerate API

raniejade:
Forgot to regenerate it after I removed the constant enum values.


godot-kotlin/godot-library/src/nativeGen/kotlin/godot/ARVRInterface.kt
```diff
@@ -185,31 +185,5 @@ open class ARVRInterface internal constructor(
...
-
-    final const val ARVR_UNKNOWN_TRACKING: Long = 3
-
-    final const val EYE_LEFT: Long = 1
-
-    final const val EYE_MONO: Long = 0
-
-    final const val EYE_RIGHT: Long = 2
-  }
+  companion object
```
chippmann, line 188: Now we have everywhere an empty companion object. Maybe not generate it when no constants should be generated?

chippmann (changes requested):

godot-kotlin/godot-library/src/nativeGen/kotlin/godot/ARVRInterface.kt
raniejade: @chippmann I'll do it in a follow up PR - let's just regenerate them for now.

raniejade:
Closing as #235 is regenerating the API already.

PR #233, opened 2020-08-12
Fix performance regression with camelToSnakeCase

raniejade:
Noticed a big performance due to `camelToSnakeCase`, due to how delegates work the delegate provider is called when an object is initialized - this means that every for every new object we call `camelToSnakeCase` `n` times, `n` being the number of signal the object has.


chippmann (approved):
Sorry for the mistake! And thanks for the fix

piiertho (approved):

